PR #41, opened 2020-02-07
Update deps and use absolute path to reduce bundle size

jiridostal:
https://github.com/RedHatInsights/frontend-components/pull/340


src/PresentationalComponents/InfoBox/InfoBox.js
```diff
@@ -1,11 +1,13 @@
 import {
-    Bullseye,
-    Split,
-    SplitItem,
     Text,
     TextContent,
     TextVariants
-} from '@patternfly/react-core';
+} from '@patternfly/react-core/dist/js/components/Text/';
+import { Bullseye } from '@patternfly/react-core/dist/js/layouts/Bullseye/';
```
katerinapat: @jiridostal does the `/ ` in the end really need it?

jiridostal: Yes I think so as it's path that points to "....../components/Text/index.js"

PR #149, opened 2020-06-04
Update tree shaking

### LINKED ISSUE #148: UI build optimizations
@karelhala is currently working optimizing the inventory component and he noticed that the build size of the patchman UI is very large even though the number of used components fairly small.

I would like to update some configuration files in this project which should considerably decrease the final build size output, improve performance, and allow @karelehala to do some more optimization in the inventory component.

### What would the changes be?
- change the way how certain modules (PF, FCE) are imported
  - either directly change import paths in code
  - or transform these imports via babel (preferred option)
- fiddle around with webpack config to split the node_modules chunk
  - there is no need to consolidate all node_modules in one massive chunk (it increases the initial entry point load and evaluation time)
  - there can be multiple vendors chunk for each lazy-loaded part of the app

cc @jiridostal

Hyperkid123:
closes #148 

**Please check the whole app as wrong imports may cause a runtime crash.** I tried to click through everything but I a not familiar with the application so I might have missed something.

### Changes
- added bundle analyzer to enable precise build debugging
  - you can run the analyzer with `npm run analyze` / `yarn analyze`
- allowed multiple vendors chunks for js files
  - before all files from node_modules were consolidated into one huge chunk
  - now all lazy-loaded modules have their own vendor chunks which can be shared between multiple modules
- manually changed import paths of FCE to use absolute paths to CJS bundles
- added babel transform imports plugin and configured it for pattern fly assets
  - this will automatically create correct import paths
  - **I strongly recommend doing the same for the FCE packages you can use the PF config as a baseline. Its very easy to forget to use correct import path**
- removed async component in favor of `React.lazy`
  - they do the same thing, no need to have two redundant implementations
  - Just an observation, I would look at how you handle the loading of lazy chunks and create some nice loaders to avoid the ugly blank screens between loads
- use pre-defined inventory component dependencies for PF core and icons

### Before
Build size of the APP using `build:prod` only JS (no css, no .map, no images):
```bash
  
                                                 js/Advisories.js   2.05 KiB       1  [emitted]               Advisories
    js/Advisories~AdvisoryyPage~InventoryPage~Register~Systems.js   47.1 KiB       0  [emitted]
                                              js/AdvisoryyPage.js   46.2 KiB       2  [emitted]                 
                                                        js/App.js   51.8 KiB       3  [emitted]             
                                              js/InventoryPage.js   1.93 KiB       4  [emitted]                   
                                                   js/Register.js  379 bytes       5  [emitted]               
                                                    js/Systems.js    6.4 KiB       6  [emitted]               
                                                     js/vendor.js    4.7 MiB       7  [emitted]        [big]  

``` 
**Around 5 MiB** total size parsed

### After 
Build size of the APP using `build:prod` only JS (no css, no .map, no images):
```
                                                     js/Advisories.js   2.05 KiB       2  [emitted]               
                                                  js/AdvisoryyPage.js   10.2 KiB       3  [emitted]            
                                                            js/App.js   1.35 MiB       4  [emitted]        [big]  
                                                  js/InventoryPage.js   8.74 KiB       5  [emitted]            
                                                       js/Register.js  379 bytes       6  [emitted]               
                                                        js/Systems.js    3.8 KiB       7  [emitted]               
                                   js/defaultVendors~AdvisoryyPage.js   37.9 KiB       8  [emitted]
js/default~Advisories~AdvisoryyPage~InventoryPage~Register~Systems.js   19.4 KiB       0  [emitted]
                                  js/default~AdvisoryyPage~Systems.js   9.54 KiB       1  [emitted]    
```
**1.44 MiB** total size parsed 

There is still one duplication of PF react-table packages, but that comes from frontend components. I will check that one after I am finished with this PR. It should fix itself in one of the future releases of frontend components.

cc @jiridostal


karelhala:
ping @jiridostal

jiridostal:
That is a lot of changes, give me a few moment to check it.

jiridostal:
Thanks for that, everything seems to work as expected!

PR #177, opened 2020-06-29
Split store when building to specific file so we have even better treeshaking

karelhala:
This PR will allow importing only store without the component itself.

ping @jiridostal


jiridostal:
Cool, thanks!

karelhala:
@jiridostal is there any way how to release new version?

PR #237, opened 2020-08-17
chore(unit tests): for api.js and DataMappers.js`

mkholjuraev:
- it seems that [fix(api) res.json() is not a fucntion error fix](https://github.com/RedHatInsights/patchman-ui/commit/277395626843a20cbed002b1d8011840926e2acb) was not a complete fix. I believe this PR has a complete fix.

- I was not able to implement some tests of functions in api.js file. They can not be tested as they call another function in the same file( [Problem definition](https://stackoverflow.com/questions/52650367/jestjs-how-to-test-function-being-called-inside-another-function) )

### DIFF
--- a/src/Utilities/api.js
+++ b/src/Utilities/api.js
@@ -25,20 +25,17 @@ export function createApiCall(
             if (contentType.indexOf('json') !== -1) {
                 throw res;
             } else {
-                throw new Promise(resolve =>
-                    resolve({
-                        errors: [
-                            { status: res.status, detail: res.statusText }
-                        ]
-                    })
-                );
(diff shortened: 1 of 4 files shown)


src/Utilities/api.js
```diff
@@ -54,7 +51,7 @@ export function createApiCall(
 }
 
 export const fetchApplicableAdvisoriesApi = params => {
-    return createApiCall('/advisories', 'get', params);
+    return createApiCall('/advisoriess', 'get', params);
```
jiridostal, line 54: Seems like a typo

PR #241, opened 2020-08-20
fix(Icons): applicable advisories icons according to mockups

mkholjuraev:
fixes: [SPM-424](https://projects.engineering.redhat.com/browse/SPM-424) and [SPM-387](https://projects.engineering.redhat.com/browse/SPM-387)

1. Icon spacing in Inventory table ( could not run Inventory locally to check if actually fixed, but I believe it is fixed)
2. Security advisories icon should be black.
3. Order of the icons should be:  [1] security advisories [2] bug fixes [3] enhancements


src/PresentationalComponents/AdvisoryType/AdvisoryType.js
```diff
@@ -1,16 +1,17 @@
...
 import React from 'react';
 import { advisoryTypes } from '../../Utilities/constants';
+import { Flex, FlexItem } from '@patternfly/react-core';
 
 const AdvisoryType = ({ type }) => {
     const advisoryType =
         advisoryTypes.find(item => item.value === type) || advisoryTypes[3];
     return (
         <React.Fragment>
-            <div className={'icon-with-label'}>
```
jiridostal, line 10: Aren't we ready to remove the class "icon-with-label" completely when using Flex component? It may be at other places too.

PR #278, opened 2020-09-08
feat(filter): Filter by upgradable on packages

jiridostal:


src/store/Reducers/SystemPackageListStore.js
```diff
@@ -2,7 +2,8 @@ import { STATUS_LOADING, STATUS_REJECTED, STATUS_RESOLVED, storeListDefaults } f
 import { changeListParams, getNewSelectedItems } from '../../Utilities/Helpers';
 import * as ActionTypes from '../ActionTypes';
 
-export const SystemPackageListStore = (state = storeListDefaults, action) => {
+let stateInit = { queryParams: { filter: { updatable: 'true' } } };
```
mkholjuraev, line 5: Just out of my curiosity, do we need string type of 'true' or bool type true

mkholjuraev (approved):

src/store/Reducers/SystemPackageListStore.js
jiridostal: as it is a value that is provided to "Radio" component by PF that accepts strings only, it is really a string that represents boolean

PR #317, opened 2020-10-23
Remove debug enable function from patch component mount

karelhala:
### No need to use `insights.chrome.enable.globalFilter();`

When navigating to patch and then to any other app it shows the global filter no matter if we hid it or not. This is caused by using `insights.chrome.enable.globalFilter();`, this function should be used only for local debugging and not stored in your code, it'll set localStorage value which enables global filter everywhere.


karelhala:
ping @mkholjuraev @jiridostal

PR #327, opened 2020-11-11
chore(optimization) some shared presentational components

mkholjuraev:
A partl of: [SPM-614](https://issues.redhat.com/browse/SPM-614)

### DIFF
--- a/src/PresentationalComponents/Snippets/AdvisoriesIcon.js
+++ b/src/PresentationalComponents/Snippets/AdvisoriesIcon.js
@@ -1,0 +1,24 @@
+import React from 'react';
+import { Flex, FlexItem, Tooltip } from '@patternfly/react-core';
+import propTypes from 'prop-types';
+
+const AdvisoriesIcon = ({ count, tooltipText, Icon }) =>(
+    <Tooltip content={tooltipText}>
+        <Flex flex={{ default: 'inlineFlex' }}>
+            <FlexItem spacer={{ default: 'spacerSm' }}>
+                <Icon/>
+            </FlexItem>
+            <FlexItem spacer={{ default: 'spacerSm' }}>
+                {count.toString()}
+            </FlexItem>
+        </Flex>
+    </Tooltip>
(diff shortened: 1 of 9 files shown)


src/PresentationalComponents/Snippets/PackagesErrataLink.js
```diff
@@ -0,0 +1,23 @@
...
+            new RegExp('\\n(?=[^\\n])', 'g'),
+            ''
+        ), 570, handlePatchLink(entityTypes.advisories, row.id, 'Read more'))}
+    </Text>
+    <ExternalLink link={`https://access.redhat.com/errata/${row.id}`}
+        text={'View packages and errata at access.redhat.com'} />
+</TextContent>);
+
+PackagesErrataLink.propTypes = {
+    row: propTypes.object
```
jiridostal, line 22: As we pass "row", can you please specify more detailed propType? That it should contain "id" and "attributes"

jiridostal: Or mabe use directly props -> id and description. Also, the name of the component does not say that it contains also description. Maybe "DescriptionWithLink"? Or something like that

src/PresentationalComponents/Snippets/AdvisoriesIcon.js
```diff
@@ -0,0 +1,24 @@
...
+            <FlexItem spacer={{ default: 'spacerSm' }}>
+                {rh.toString()}
+            </FlexItem>
+        </Flex>
+    </Tooltip>
+);
+
+AdvisoriesIcon.propTypes = {
+    Icon: propTypes.element,
+    rh: propTypes.any,
```
jiridostal, line 20: "rh" does not say much, I think that something like "number" or "count" would be more explanatory. The same for "content", "tooltipText" would be better IMO

jiridostal (changes requested):
Left a few comments, other than that it looks great!

PR #355, opened 2020-12-09
fix(treeshaking): Improve treeshaking of imported patch detail

karelhala:
### Imported patch detail bloats the app bundle

When importing patch detail the final build is bloated by approx 4MB, this PR fixes such issue by improving how PF and FEC files are imported. When finally imported the build increased only by 0.4MB.


karelhala:
ping @mkholjuraev @jiridostal

jiridostal:
Looks great! Thanks

PR #381, opened 2021-01-07
feat(Packages):  sort by updatable packages by default

mkholjuraev:
resolves: https://issues.redhat.com/browse/SPM-651


src/Messages.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable sort-keys */
```
jiridostal, line 1: Hehe, don't disable this one :-)

jiridostal (changes requested):

src/Messages.js
mkholjuraev: Actually, this line was supposed to be deleted ::)

jiridostal:
Rebase needed :-(

PR #385, opened 2021-01-11
fix(InventoryEntitiesReducer): remove extra store utilization

mkholjuraev:
fixes: 'You may not call store.getState() while the reducer is executing. The reducer has already received the state as an argument. Pass it down from the top reducer instead of reading it from the store.' issue related to fetching an already existing store inside InventoryEntitiesReducer.  The issue occurred after updating frontend-components-utilities to 2.2.9.


jiridostal:
Closing as it was solved in https://github.com/RedHatInsights/patchman-ui/pull/389

PR #386, opened 2021-01-11
chore(dependabot): extend the update checking period

mkholjuraev:
I would suggest extending the update checking period from _live_ to _weekly_ so that we do not have every version updates which ends up with a lot of PRs opened.


jiridostal:
Great idea !

PR #393, opened 2021-01-14
feat(Advisories): display cve info

mkholjuraev:
Resolves: https://issues.redhat.com/browse/SPM-675


jiridostal:
Looks good! Only the sorting does not seem to work.

PR #394, opened 2021-01-18
Update app build to be compatible with chrome 2.0

Hyperkid123:
This PR makes the Patchman UI compatible with Chrome 2.0. The application will still work with current chrome, but in order to deploy chrome 2.0, all apps that are using the inventory components, have to be migrated first due to a change in the inventory distribution system. This PR is slightly bigger than usual due to some old node packages which are no longer supported in current version of webpack.

### Changes
- update dependencies
  - there were a lot of dependencies that are no longer compatible with webpack 5. I took the liberty of updating all dev-dependencies and FEC dependencies
  - I have also noticed that the lock file was generated with npm@7+. I just want to point out that most of the FEC dependencies  and PF react-core (next release fixes it) are not yet compatible with npm 7 so I reccomend would avoid using it for now
- Use updated FEC config to create chrome 2.0 required build
- Use chrome 2.0 compatible inventory components
  - current system does not work with chrome 2.0 inventory which is using federated modules to distribute Inventory components
- Fixed async tests in AdvisoryDetail components
  -  async components were not waiting for the async lifecycle to run which lead to false positive/negative results based on in which order are those tests run.
  - Now tests cases are using unique instances of the component where necessary and are waiting for the instance async lifecycle to complete to get the correct test results

cc @jiridostal


jiridostal:
@Hyperkid123 Unfortunately, there is an issue with system detail page. It triggers some error that caused the whole component tree to be recreated from error boundary and gets in a loop, causing a stuck browser... :-(

Hyperkid123:
@jiridostal it is actually supposed to load the ErrorComponent. That is a fallback until chrome 2.0 is available and it uses the current inventory. It's by design.

However, the loading loop is not. I will look at why is happening.

Hyperkid123:
@jiridostal turns out that I've made a silly mistake. I've assigned the `onLoad` function prop to the wrong element. Instead of the `InventoryDetailHead` to your `Header` component.

I will publish the fix and also deal with the conflicts.

PR #400, opened 2021-01-21
All tables should be compactfor ocnsistency

jiridostal:
image
image
image
image


terezanovotna:
Looks great! Thanks for changing it up so quickly. Feel free to tag me even here on gh when we have any UI updates

mkholjuraev (approved):

PR #404, opened 2021-01-26
bug(StatusFilter): clear filters on unmount, solve 2x upgradable label

mkholjuraev:
Resolves: https://issues.redhat.com/browse/SPM-651 and https://issues.redhat.com/browse/SPM-697


mkholjuraev:
Better to merge after https://issues.redhat.com/browse/SPM-703 is resolved.

mkholjuraev:
The final solution is kind of hackish. SystemPackages and packages pages need to send their own filter object name and types.
I think it would be better if we have a short discussion about it later. 

In this commit, I removed clearing PackagesListStore as we have discussed.

jiridostal:
Package detail page does not load with an error:
image

Also, isn't it nicer to have two separate filters than merging these two cases into one?

PR #413, opened 2021-01-28
fix(notifications): Display notification when remediation is created

jiridostal:


mkholjuraev (approved):
I am wondering if we need to refresh the page and exclude remediated advisory from the table after remediation is succeded?

PR #434, opened 2021-02-12
feat(Packages): change remediation identifier to patch-package

mkholjuraev:
Resolves: https://issues.redhat.com/browse/SPM-737

- I think we have only 2 usages of remediation (remediation advisories, and remediation packages). Thus I did want to touch on existing, already working code. Instead created a special case for remediating packages.
- some unused code is removed from Packages component


jiridostal (changes requested):
Isn't is better to pass a constant to the remediationProvider, instead a boolean? We can create two constants - for package and for advisories so the developer can pick what he wants (also for potential future remediation types)

jiridostal:
Nice.

PR #437, opened 2021-02-15
fix(select): after filering

mkholjuraev:
resolves: https://issues.redhat.com/browse/SPM-732


jiridostal:
https://github.com/RedHatInsights/frontend-components/pull/962

PR #444, opened 2021-02-18
Remove insights route as it can cause page break

karelhala:
### Insights route should not be used

Since we are in SPA now InsightsRoute is redundant and only the `Route` from `react-router-dom` can be used.

### DIFF
--- a/src/Routes.js
+++ b/src/Routes.js
@@ -85,30 +85,6 @@ export const paths = {
     }
 };
 
-const InsightsRoute = ({ component: Component, rootClass, ...rest }) => {
-    const root = document.getElementById('root');
-    /**
-     * @deprecated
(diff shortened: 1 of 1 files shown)


src/Routes.js
```diff
@@ -137,35 +113,32 @@ export const Routes = (props) => {
...
                     path={paths.packages.to}
                     component={PackagsPage}
-                    rootClass="Patchman"
                 />
-                <InsightsRoute
+                <Route
                     exact
                     path={paths.register.to}
                     component={RegisterPage}
-                    rootClass="Patchman"
```
Hyperkid123, line 166: @karelhala can you check if this className is used in some of the scss files?

karelhala: Yup, checked and I couldn't find it anywhere.

PR #468, opened 2021-03-17
fix(SonarCube): issues are resolved

mkholjuraev:
resolves: https://issues.redhat.com/browse/SPM-792

SonarCube reliability bugs are fixed. Maintainability bugs will be fixed in the next PR

### DIFF
--- a/src/SmartComponents/SystemDetail/SystemDetail.js
+++ b/src/SmartComponents/SystemDetail/SystemDetail.js
@@ -13,7 +13,7 @@ const SystemDetail = ({ isInventoryApp }) => {
     const [activeTabKey, setActiveTabKey] = React.useState(0);
     const [areTabsHidden, setTabsHidden] = React.useState(false);
 
-    const entity = useSelector(({ entityDetails }) => entityDetails.entity && entityDetails.entity || {});
+    const entity = useSelector(({ entityDetails }) => entityDetails.entity || {});
     const onTabSelect = (event, id) => {
         setActiveTabKey(id);
     };
--- a/src/Utilities/Helpers.js
+++ b/src/Utilities/Helpers.js
@@ -330,8 +330,8 @@ export function sortCves(cves, index, direction) {
 
     const sortedCves = cves.sort(
         ({ cells: aCells }, { cells: bCells }) => {
-            const aCell = aCells[index].value && aCells[index].value || aCells[index].title;
-            const bCell = bCells[index].value && bCells[index].value || bCells[index].title;
+            const aCell = aCells[index].value || aCells[index].title;
+            const bCell = bCells[index].value || bCells[index].title;
 
             const stringA = aCell.toString().toUpperCase();
             const stringB = bCell.toString().toUpperCase();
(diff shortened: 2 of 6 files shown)


src/Utilities/Helpers.js
```diff
@@ -15,6 +15,10 @@ import { Flex, FlexItem, Tooltip } from '@patternfly/react-core';
 import messages from '../Messages';
 import { intl } from './IntlProvider';
 
+export const isSet = (variable) => (
```
jiridostal, line 18: Is there any reason why we need it?

src/SmartComponents/SystemDetail/SystemDetail.js
```diff
@@ -8,12 +8,13 @@ import { intl } from '../../Utilities/IntlProvider';
...
 import { NoSystemData } from '../../PresentationalComponents/Snippets/NoSystemData';
 import propTypes from 'prop-types';
+import { isSet } from '../../Utilities/Helpers';
 
 const SystemDetail = ({ isInventoryApp }) => {
     const [activeTabKey, setActiveTabKey] = React.useState(0);
     const [areTabsHidden, setTabsHidden] = React.useState(false);
 
-    const entity = useSelector(({ entityDetails }) => entityDetails.entity && entityDetails.entity || {});
+    const entity = useSelector(({ entityDetails }) => isSet(entityDetails.entity) && entityDetails.entity || {});
```
jiridostal, line 17:
```suggestion
    const entity = useSelector(({ entityDetails }) => entityDetails.entity || {});
```
Why so complicated? It should work the same

src/Utilities/Helpers.js
```diff
@@ -330,8 +334,8 @@ export function sortCves(cves, index, direction) {
 
     const sortedCves = cves.sort(
         ({ cells: aCells }, { cells: bCells }) => {
-            const aCell = aCells[index].value && aCells[index].value || aCells[index].title;
-            const bCell = bCells[index].value && bCells[index].value || bCells[index].title;
+            const aCell = isSet(aCells[index].value) && aCells[index].value || aCells[index].title;
```
jiridostal, line 337:
```suggestion
            const aCell = aCells[index].value || aCells[index].title;
```
Also for other occurrences

jiridostal (changes requested):
isSet function is not needed at all. Some resources:

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_AND
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

src/Utilities/Helpers.js
mkholjuraev: I also think that this is not needed. But the issue itself is weird https://sonarqube.corp.redhat.com/project/issues?id=insights%3Apatchman-ui&issues=AXfUchTERQz83K5m01H6&open=AXfUchTERQz83K5m01H6

PR #469, opened 2021-03-18
SPM-755 Add bulk remediation for systems and advisories

jiridostal:
This PR implement the Patch side of bulk remediation.
**There is a bug that prevents the remediation from creating - the wizard goes to the last step and than fails, platform is working on it**

The purpose is to demonstrate the behavior on Patch side.

### DIFF
--- a/src/Utilities/Helpers.js
+++ b/src/Utilities/Helpers.js
@@ -172,7 +172,6 @@ export async function remediationProviderWithPairs(issues, createPairs, transfor
     if (issues) {
         const pairsCreated = await createPairs(issues);
         const res = transformFunc(pairsCreated);
-        console.log(await res);
         return await res;
     }
     else {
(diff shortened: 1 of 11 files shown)


mkholjuraev (changes requested):
I am afraid the PatchRemediationButton component is missing from the PR commit.

jiridostal:
Fixed!

src/Utilities/Helpers.js
```diff
@@ -154,6 +168,20 @@ export const remediationProvider = (issues, systems, remediationIdentifier) => {
         : false;
 };
 
+export async function remediationProviderWithPairs(issues, createPairs, transformFunc) {
+    if (issues) {
+        const pairsCreated = await createPairs(issues);
+        const res = transformFunc(pairsCreated);
+        console.log(await res);
```
mkholjuraev, line 175: I believe this needs to be deleted :)

mkholjuraev (commented):
It looks that 2 remediation modals are being mounted in the Systems page, when I close one model, the other is displayed. Otherwise working as described.

PR #477, opened 2021-03-25
SPM-769  feat(git commits): use commitizer to enable conventionam commit naming

mkholjuraev:
This is a test PR from using commitizer for the ticket https://issues.redhat.com/browse/SPM-769

[cz-cli](https://github.com/commitizen) has several steps to finish the commit e.g type, scope, footer .... I thought some of them are unnecessaryy, thus I used [cz-customizable](https://github.com/leonardoanalista/cz-customizable) with custom config which can be improved according to needs later.

### DIFF
--- a/.cz-config.js
+++ b/.cz-config.js
@@ -1,0 +1,59 @@
+module.exports = {
+    types: [
+        { value: 'feat', name: 'feat:     A new feature' },
+        { value: 'fix', name: 'fix:      A bug fix' },
+        { value: 'WIP', name: 'WIP:      Work in progress' },
+        {
+            value: 'chore',
+            name:
+                'chore:    Changes to the build process or auxiliary tools\n            and libraries such as documentation generation',
+        },
+        { value: 'docs', name: 'docs:     Documentation only changes' },
+        {
+            value: 'style',
+            name:
+                'style:    Changes that do not affect the meaning of the code\n            (white-space, formatting, missing semi-colons, etc)',
+        },
+        {
+            value: 'refactor',
+            name: 'refactor: A code change that neither fixes a bug nor adds a feature',
+        },
+        {
+            value: 'perf',
+            name: 'perf:     A code change that improves performance',
+        },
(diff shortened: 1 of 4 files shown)


.cz-config.js
```diff
@@ -0,0 +1,58 @@
...
+                'chore:    Changes to the build process or auxiliary tools\n            and libraries such as documentation generation',
+        },
+        { value: 'docs', name: 'docs:     Documentation only changes' },
+        {
+            value: 'style',
+            name:
+                'style:    Changes that do not affect the meaning of the code\n            (white-space, formatting, missing semi-colons, etc)',
+        },
+        {
+            value: 'refactor',
```
jiridostal, line 18: Maybe we should add this rule to semantic-release, so it triggers a "patch"(fix) build.

.cz-config.js
```diff
@@ -0,0 +1,58 @@
...
+            name: 'refactor: A code change that neither fixes a bug nor adds a feature',
+        },
+        {
+            value: 'perf',
```
jiridostal, line 22: Same as above, let's treat this as a fix in semantic release

.cz-config.js
```diff
@@ -0,0 +1,58 @@
...
+
+    scopes: [],
+
+    allowTicketNumber: true,
+    isTicketNumberRequired: true,
+    ticketNumberPrefix: 'SPM-',
+    ticketNumberRegExp: '',
+
+    // override the messages, defaults are as follows
+    messages: {
```
jiridostal, line 38: The "Ticket number" message should say that "SPM" is added automatically

jiridostal (changes requested):
Looks and works really great, just some minor improvements!

.cz-config.js
mkholjuraev: I see there is 'refactor' in semantic release config https://github.com/RedHatInsights/patchman-ui/blob/master/package.json#L137. Am I looking at the right place? I am not very experienced with semantic release config :)

PR #484, opened 2021-03-31
Wrap inventory detail with wrapper

karelhala:
### Use inventory wrapper

Inventory detail consists of multiple components, and they require direct access to redux. This PR changes the structure of system detail in order to merge inventory redux with patch's one.


mkholjuraev:
@karelhala when I run this locally, I still see the issue. Should I do something more? And not only advisory detail, but also systems page also not loading

mkholjuraev (approved):

PR #502, opened 2021-04-14
fix(bulk select): SPM-848 make bulk select consistence across tables

mkholjuraev:


jiridostal:
If you go to "/patch/packages", check that there is a checkbox and is in an incorrect state (I think it should not be there at all)

PR #527, opened 2021-05-06
feat(system detail): use federated modules to share system detail

karelhala:
### Use federated modules to share system detail

Since we can now use federated modules in every environment let's share the system detail with it. This way we don't have to maintain the version of it in all environments, it's the same for all users on one environment. If a bug is found the fix is deployed just in this application and propagated to inventory via federetad modules.

### Usage
```JSX
import React, { useContext } from 'react';
import AsyncComponent from '@redhat-cloud-services/frontend-components/AsyncComponent';
import { RegistryContext } from '../../store';
import { useRouteMatch } from 'react-router-dom';

const PatchTab = () => (
  <AsyncComponent
      appName="patch"
      module="./SystemDetail"
      getRegistry={useContext(RegistryContext).getRegistry}
  />);
```


karelhala:
ping @mkholjuraev @jiridostal

mkholjuraev (approved):
The PR looks good to me, but I was not able to run it locally and try it out. Inventory app breaks with the error: **Error while loading component!**. 

image

karelhala:
@mkholjuraev you'll have to run patch app on port 8002 and inventory on port 8003 and change your config to serve `inventory`. I'll add this to your spandx file

mkholjuraev (approved):

PR #544, opened 2021-05-20
fix(SelectedRows): SPM-924 filter out previously deselected rows

mkholjuraev:


mkholjuraev:
Merging the PR as it would be good if we can add this to the release.

PR #557, opened 2021-06-02
Integrate handle refresh into getEntities.

Hyperkid123:
cc @mkholjuraev 

This PR fixes an issue when the inventory table hangs in a loading state.

The problem was caused by duplicate API calls by the `getEntities` and `onRefresh` callbacks. Where possible, I've integrated the onRefresh into the `getEntities` callback which will prevent double calls and hanging loading state in the entities reducer.

@mkholjuraev Please double-check that nothing broke with this change.


mkholjuraev (approved):
Thanks for the PR, I have faced a small issue with 'Clear filters'. But it is an easy to fix I believe.

PR #590, opened 2021-07-09
chore(PersistantParams): SPM- 1006 make all pages consistant to apply url param…

mkholjuraev:
…s on load

### DIFF
--- a/src/SmartComponents/Packages/Packages.js
+++ b/src/SmartComponents/Packages/Packages.js
@@ -11,12 +11,15 @@ import { changePackagesListParams, fetchPackagesAction } from '../../store/Actio
 import { exportPackagesCSV, exportPackagesJSON } from '../../Utilities/api';
 import { packagesListDefaultFilters } from '../../Utilities/constants';
 import { createPackagesRows } from '../../Utilities/DataMappers';
-import { createSortBy } from '../../Utilities/Helpers';
+import { createSortBy, decodeQueryparams, encodeURLParams } from '../../Utilities/Helpers';
 import { setPageTitle, useOnExport, usePerPageSelect, useSetPage, useSortColumn } from '../../Utilities/Hooks';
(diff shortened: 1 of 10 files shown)


src/Utilities/Helpers.js
```diff
@@ -19,6 +19,11 @@ import {
 import { intl } from './IntlProvider';
 import { packageSystemsColumns } from '../SmartComponents/Systems/SystemsListAssets';
 
+export const deleteUnsetObjectValues = (object) => {
```
jiridostal, line 22: This is very similar to "filterSelectedRowIDs". We should think about consolidating naming of these utils, we have quite a lot of them and I'm not sure if we need all of them

src/Utilities/Helpers.js
```diff
@@ -19,6 +19,11 @@ import {
...
+    Object.keys(object).forEach(key => object[key] === undefined && delete object[key]);
```
jiridostal, line 23: `delete object[key]` modifies the original object, it might be more useful if the original object remains unchanged (create a copy and delete it there). Just an idea that would make our debugging easier.

jiridostal (changes requested):
Some notes in the code... but works perfectly!

src/SmartComponents/Packages/Packages.js
```diff
@@ -35,10 +38,17 @@ const Packages = () => {
     );
 
     React.useEffect(() => {
-        dispatch(fetchPackagesAction(queryParams));
+        if (firstMount) {
```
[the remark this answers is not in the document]
mkholjuraev: hmm, I think we have to use this approach which is similar to the Advisories page. Details pages use the Inventory component which enables us to populate starting redux state with the decoded URL params. But yes, it would be great if we can come up with more optimal solution.

PR #594, opened 2021-07-14
feat(Version filter): SPM-914 version filter is finished

mkholjuraev:


src/PresentationalComponents/Filters/VersionFilter.js
```diff
@@ -0,0 +1,78 @@
...
+    const onSelect = (_, selection) => {
+        const existingVersions = isSet && filter.installed_evra !== '' && filter.installed_evra.concat(',') || '';
+        apply({ filter: { installed_evra: `${existingVersions}${selection}` } });
+    };
+
+    const clearSelection = () => {
+        setOpen(false);
+    };
+
+    const simulateNetworkCall = callback => {
```
jiridostal, line 26: We can remove this :-)

src/PresentationalComponents/Filters/VersionFilter.js
```diff
@@ -0,0 +1,78 @@
...
+        setTimeout(callback, 1000);
+    };
+
+    const onViewMoreClick = () => {
+        setLoading(true);
```
jiridostal, line 31: We don't have to simulate loading

jiridostal:
Looks great! A few notes:

- When version is selected from the list, it can not be unchecked
- I don't like the filter name "Version filter", maybe just "Version" is better
- THe "X" icon in circle to remove the filter does not do anything
- When I click "View more" I would expect to load the rest of the list, but it loads only a couple next items, maybe UX needed
- Can not we show at least 10 items when first opening the dropdown?

mkholjuraev:
- loading effect is removed
- unsetting a version is enabled
- X button is removed to be consistent with other filters. We have already this feature from FEC
- 10 versions are listed by default
- Version filter is changed to => Version

PR #595, opened 2021-07-15
fix(Inventory): SPM-1019 unwanted hosts are removed

mkholjuraev:


src/Utilities/Hooks.js
```diff
@@ -120,7 +120,17 @@ export const useOnSelect = (rawData, selectedRows, fetchAllData, selectRows,
     const onSelect = React.useCallback((event, selected, rowId) => {
         const createSelectedRow = (rawData, toSelect = []) => {
             rawData.forEach((row) => {
-                toSelect.push(
+                const {
```
jiridostal, line 123: I think this approach is not very flexible ... Can not we pass another function to the onSelect, let's say 'isRowDisabled' that would take the row and evaluate according to the provided function? In this case we would be able to use different condition from different pages. Generally I don't like the fact that we have that many constants in a hook

PR #596, opened 2021-07-17
fix(CVE-2021-33623) SPM-970 is remediated

mkholjuraev:
The `trim-newlines` package before 3.0.1 and 4.x before 4.0.1 for Node.js has an issue related to regular expression denial-of-service (ReDoS) for the .end() method.

By updating node-sass to 6.0.1 we are using only  `trim-newlines@3.0.1`


mkholjuraev:
It is painful to find out what is the problem. @jiridostal  Let's merge this, if you think it is safe

PR #598, opened 2021-07-21
feat(NoAccess): SPM-968 use shared not-connected component

mkholjuraev:


mkholjuraev:
I have no idea why tests are not passing

jiridostal:
Delete package-lock and install again, it should fix the issue.

mkholjuraev:
Here also, It is painful to find out what is the problem. @jiridostal Let's merge this if you think it is safe

PR #609, opened 2021-07-28
fix(Inventory): problems after rebase are solved

mkholjuraev:
The issues after rebase:
1. `filterSelectedRowIDs` function in Helpers.js was renamed to `removeUndefinedObjectKeys`. But this change was got lost during rebase and it was not reflected in Inventory Components.
2. `useEffect` in PackageSystems page was changed to async function and it broke the implementation of `clearInventoryReducer`. Thus if you select a system on the Package Systems page it was present also in Systems Page. 
3. If you select All systems, all of the systems on the first page is selected even if it was selected. It was because we did not call the `inventoryModifier` reducer to remap with updated systems selected.
4. The last problem that occurred was the incorrect `useSelector` for selected rows in all Inventory pages. Previously `selectedRows` was a part of the relevant reducer for the component. But after rebase, it is now stored in `entities` reducer


jiridostal:
did not try the change, but let's merge it and observe in CI.

PR #610, opened 2021-07-29
fix(Clear Filters): SPM-1020 redundant API call is removed

mkholjuraev:


jiridostal:
Seems like there are some styling issues 
image

PR #631, opened 2021-08-17
SPM-1113 Fix inconsistent spacing on toolbars/tables

jiridostal:


mkholjuraev (approved):
I am wondering if you have a special prettifier that sorts file includes alphabetically. All the imports are sorted and it is good.

jiridostal:
I think it's a plugin called 'organize-imports' :-)

PR #633, opened 2021-08-18
fix(Filter Chips): SPM-1095 make search filter chip text consistent w…

mkholjuraev:
…ith search filter title

I am not sure if we really want this change. As search can be made in multiple columns and naming filter chip with 'System' or 'Advisory' seems confusing to me.


jiridostal:
Merged. We can collect feedback and change.

PR #637, opened 2021-08-20
feat(StatusReport): SPM-1043 create System status report

mkholjuraev:
There is a spacing difference between the top and bottom of the status bar. I liked that :). Because the status bar and the table are one group and it makes sense to put it a bit far from the header.

image


jiridostal:
Looks great! A few notes:

- Can we add a spinner or indicate that the number is loading?
- Numbers should be clickable - not only blue - on click it should filter the list

jiridostal:
The title should not be "Most impactful advisories". :-)
The space beween the icon and number is too big, make it smaller please.
Other than that, instant merge!

mkholjuraev:
Yeah, I have to live with stupid mistakes :)

I have noticed that in the mockup there is not a title at all. https://marvelapp.com/prototype/15116dj4/screen/80040041

This is how UI looks now

image

PR #653, opened 2021-09-20
feat(AdvisoriesReport): SPM-1167 add most impactfull advisory info

mkholjuraev:
I have a few suggestions already added into the UI design: 
1. Move advisory name into the title position of the Patternfly card
2. move the number of applicable systems into a separate block GridItem  so that it is not split into 2 lines
3. Move os name to first column Grid item elements so that we have a balanced number of info labels into 2 sides.

image

I have not put a loader, because I am not sure if we have always more than 4 advisories in the database. If we always have and we know what will be shown, the loader can be put always

### DIFF
--- a/src/Messages.js
+++ b/src/Messages.js
@@ -11,6 +11,11 @@ export default defineMessages({
         description: 'regsiter page title',
         defaultMessage: 'Patch'
     },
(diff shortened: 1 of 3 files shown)


mkholjuraev:
Suggestions are approved UX!

PR #684, opened 2021-11-02
feat(Tags): SPM-1178 add tags column, tag filter

mkholjuraev:


mkholjuraev:
The global tag filter and local tag filter in the inventory table are unified, giving the privilege to the local tag.

jiridostal:
obrazek

I see the Tags column is a bit misaligned. Btw is change of the order of columns intended?

jiridostal:
obrazek

Here I see there is only one Tag available per row, but when I click it I see 6 tags

PR #689, opened 2021-12-01
fix(Last Seen): fix invalid date in last seen column

mkholjuraev:
Fixes: https://ansible.slack.com/archives/CPS5FH07Q/p1638342210212600


jiridostal:
@mkholjuraev looks and works good, but tests are failing

PR #703, opened 2022-01-14
SPM-1287: make system status card respect only tags, os, systemPlatform

mkholjuraev:
GlobalFilterReducer usage is enhanced together with the ticket expectation.


src/store/Reducers/GlobalFilterStore.js
```diff
@@ -1,26 +1,34 @@
...
 
-export const GlobalFilterStore = (state = {}, action) => {
+const initialState = {
+    selectedGlobalTags: [],
+    selectedTags: [],
+    systemProfile: undefined
+};
+
+export const GlobalFilterStore = (state = initialState, action) => {
+    let newState = { ...state };
```
bastilian, line 10: This isn't really needed, or am I missing something?

mkholjuraev: Hmm, maybe it is really not needed. But the intention of this line was to avoid mutation of the initial state accidentally by creating a copy of the state inside reducer. Do you think that it makes sense or should I remove it. We are using this in all reducers.

bastilian: Hm. I don't think using `state` directly below and spread it would mutate it or `initialState`. Unless I'm missing something. 

I think, and I'm not sure 100%, that a case where the `initialState` could be mutated is when this line would just be `let newState = state`, but initialState is a `const`, so something would say that it's not possible. And even with that line there would need to be and action that tried to assign or mutate `newState`.

mkholjuraev: Good, I have adopted your suggestion.

src/SmartComponents/SystemPackages/__snapshots__/SystemPackages.test.js.snap
```diff
@@ -6408,7 +6414,7 @@ exports[`SystemPackages.js Should match the snapshots 1`] = `
                                       getMenuRef={null}
                                       id="pagination-options-menu-bottom-toggle-1"
                                       isActive={false}
-                                      isDisabled={true}
+                                      isDisabled={0}
```
bastilian, line 6417: This looks a bit fishy to me. Is that right?

mkholjuraev: I updated the snapshots. Maybe it is due to some changes in the dependencies. But I would not care much about this :)

bastilian: OK. :D

bastilian (approved):
Works for me! Thank you @mkholjuraev!

PR #713, opened 2022-02-02
feat(SPM-1306): create a patch set wizard

mkholjuraev:
We have incomplete UI mockup designs. I would like to improve the wizard once the mockups are complete.


src/SmartComponents/PatchSetWizard/PatchSetWizard.js
```diff
@@ -0,0 +1,133 @@
...
+                                        state={wizardState}
+                                    />
+                                ),
+                                isFinishedStep: true
+                            }
+                        ]}
+                        onClose={handleWizardClose}
+                    />
+                </Modal>
+            ) : null}
```
bastilian, line 121: Is there a specific reason why this is not just one `{ ... ? ... : ... }` ?

src/SmartComponents/PatchSetWizard/steps/RequestProgress.js
```diff
@@ -0,0 +1,94 @@
...
+                    : percent === 100
+                        ? 'Patch Set creation successfull'
+                        : 'Patch Set creation in progress'}
+            </Title>
+            <EmptyStateBody>
+                <Grid hasGutter>
+                    <GridItem>
+                        <ProgressBar percent={percent} failed={failed} />
+                    </GridItem>
+                    {(percent !== 100 && !failed) && (
```
bastilian, line 58: Could you not just wrap the second `GridItem` with the condition(s)?

src/SmartComponents/PatchSetWizard/InputFields/DescriptionField.js
```diff
@@ -0,0 +1,29 @@
+import React, { useState } from 'react';
+import { FormGroup, TextInput } from '@patternfly/react-core';
+import useFormApi from '@data-driven-forms/react-form-renderer/use-form-api';
+import useFieldApi from '@data-driven-forms/react-form-renderer/use-field-api';
+
+const DescriptionField = (props) => {
+    const { input } = useFieldApi(props);
+    const formOptions = useFormApi();
+    const values = formOptions.getState()?.values;
```
bastilian, line 9: Is it not possible to use `props.value(s)`, like in the second example here: https://data-driven-forms.org/hooks/use-field-api#usageofinput

src/store/Reducers/PatchSetsReducer.js
```diff
@@ -0,0 +1,25 @@
+import * as ActionTypes from '../ActionTypes';
+
+// Initial State. It should not include page and perPage to persist them dynamically
```
bastilian, line 3: > It should not include page and perPage to persist them dynamically

... ähm... but it does include page and perPage? :D

bastilian (approved):
Looks good to me! And works as expected! Thank you @mkholjuraev!

mkholjuraev:
@bastilian Thanks for the review. I will adopt your suggestions and merge the PR

src/SmartComponents/PatchSetWizard/steps/RequestProgress.js
mkholjuraev: We should display different buttons also. Thus I think it makes sense to have a separate GridItem

mkholjuraev: Anyways, these lines will be most probably changed when the designs are ready for the last step.

src/SmartComponents/PatchSetWizard/InputFields/DescriptionField.js
mkholjuraev: I am not sure if the props would always have the value when user comes back to the input from next step.

console.log(props) does not list the value in that ^ senario.

image

PR #718, opened 2022-02-21
feat(SPM-1305): create patch set table

mkholjuraev:
This enables  to view Patch set list. Buttons, actions will be integrated in the next ticket https://issues.redhat.com/browse/SPM-1380


src/Utilities/DataMappers.js
```diff
@@ -299,3 +300,35 @@ export const createCvesRows = (rows) => {
...
+                key: row.id,
+                selected: selectedRows[row.id] !== undefined,
+                cells: [
+                    { title: row.name },
+                    { title: row.systems }
+                ]
+            };
+        });
+
+    return data.length > 0 && data ||
```
leSamo, line 320: According to statement above, `data` can possibly be `false` when rows are `false` or nullish, which can produce error when trying to access `length`, you probably want to add fallback above:
`data = rows ? rows.map(... : []`

src/Utilities/DataMappers.js
```diff
@@ -299,3 +300,35 @@ export const createCvesRows = (rows) => {
...
+        [
+            {
+                heightAuto: true,
+                cells: [
+                    {
+                        props: { colSpan: 6 },
+                        title: (filters.seach || Object.keys(filters.filter).length)
+                            && <EmptyPatchSetList/>
+                            || <NoPatchSetList/>
```
leSamo, line 329: You can use ternary operator (? and :) which is clearer then && and ||

src/Utilities/DataMappers.js
mkholjuraev: We use && and || on across Patch UI code. Because, we do not have to explicitly tell code to return undefined when we want to only check AND condition. 
E.G:
`const selectedCount = selectedRows && arrayFromObj(selectedRows).length;` [TableView: line 62](https://github.com/RedHatInsights/patchman-ui/blob/master/src/PresentationalComponents/TableView/TableView.js#L62)

`const advisoryType =  advisoryTypes.find(item => item.value === type) || advisoryTypes[3];` [AdvisoryType: line 7](https://github.com/RedHatInsights/patchman-ui/blob/master/src/PresentationalComponents/AdvisoryType/AdvisoryType.js#L7).

src/store/Reducers/PatchSetsReducer.js
```diff
@@ -0,0 +1,53 @@
+import * as ActionTypes from '../ActionTypes';
+import {
+    changeFilters,
+    selectRows,
+    fetchPending,
+    fetchRejected
+} from './HelperReducers';
+
+// Initial State. It should not include page and perPage to persist them dynamically
+export const initialState = {
```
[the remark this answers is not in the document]
bastilian: `Object.freeze` only does a shallow freeze, meaning it only freezes the immediate properties, but not properties of objects assigned to the properties. For example, while it would not be possible to do `initialState. queryParams = ...` it is possible to do `initialState.queryParams.page = 1`.

PR #721, opened 2022-02-23
fix(InventoryDetail): Changed Alert to inline

AsToNlele:
This ticket is in a wrong project: ADVISOR-1964

New look
image


MichalSajdik (approved):

leSamo:
I believe there should also be a md (16px) spacer above the alert, now it looks really crammed

AsToNlele:
You are right, now its looks much better
image

leSamo (approved):
Perfect, thank you for the additional fix and for providing pictures in your PR!

One tiny nitpick: instead of `variant="info"` you could use `variant={AlertVariant.info}` to make it more clear, but this is just a heads up into the future :grinning:

PR #740, opened 2022-03-04
fix(SPM-1381): Pagination dropdown disabled when table empty

AsToNlele:
Should be working in all parts of Patchman
image


mkholjuraev (approved):
Looks good to me! Thanks for the work.

PR #741, opened 2022-03-04
fix(PatchSetWizard): separate request to update and create set

mkholjuraev:
We need to send  distinct requests for updating and creating a patch set according to the changed [API docs](https://console.stage.redhat.com/beta/docs/api/patch#operations-default-createBaseline). The API doc is missing description field on both endpoints:

1. For creating a patch set (**/api/patch/v1/baselines**): 
  `{
    "config": {
      "to_time": "2022-12-31T12:00:00-04:00"
    },
    "inventory_ids": [
      "string"
    ],
    "name": "string"
  }` 
2. For updating a patch set (**/api/patch/v1/baselines/{baseline_id}**):
```
  {
    "config": {
      "to_time": "2022-12-31T12:00:00-04:00"
    },
    "inventory_ids": {
      "additionalProp1": true,
      "additionalProp2": true,
      "additionalProp3": true
    },
    "name": "my-changed-baseline-name"
  }
```

### DIFF
--- a/src/SmartComponents/PatchSetWizard/InputFields/SelectExistingSets.js
+++ b/src/SmartComponents/PatchSetWizard/InputFields/SelectExistingSets.js
@@ -20,7 +20,10 @@ const SelectExistingSets = ({ patchSets, setSelectedPatchSet, selectedSets, syst
         setSelectedPatchSet(selected);
 
         const selectedSet = patchSets.filter(set => set.name === selected);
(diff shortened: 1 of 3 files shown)


mkholjuraev:
I would like include this PR for my current patch set wizard integration work. Merging this as it will not be pushed to prod untill Q2 and can be tested in the next PRs

PR #749, opened 2022-03-11
feat(PatchSet): put patch set work under feature flag

mkholjuraev:
This PR is intended to put patch set work under a feature flag. This feature is inteded to be released into prod in Q2. 

**activateFeatureFlag** is used in App.js once to enable flags using URL search params and environment substring. The function calculates what features are enabled and stores them in localStorage. Later **checkEnabledFeature** can be used in any component down in the DOM to check a single feature. 

 All features are by default enabled in development and stage environments. So that we do not have to enter url search param everytime if we want a certain feature. I do not see any reason to hide a feature in these environments.

### DIFF
--- a/src/Utilities/Helpers.js
+++ b/src/Utilities/Helpers.js
@@ -6,8 +6,7 @@ import {
     SecurityIcon
 } from '@patternfly/react-icons';
 import { SortByDirection } from '@patternfly/react-table/dist/js';
-import findIndex from 'lodash/findIndex';
-import flatten from 'lodash/findIndex';
+import { findIndex, flatten } from 'lodash';
 import qs from 'query-string';
 import React from 'react';
 import LinesEllipsis from 'react-lines-ellipsis';
@@ -19,7 +18,9 @@ import {
     advisorySeverities,
     compoundSortValues,
     filterCategories,
-    multiValueFilters
+    multiValueFilters,
+    featureFlags,
+    environments
 } from './constants';
 import { intl } from './IntlProvider';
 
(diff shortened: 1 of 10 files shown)


bastilian:
@mkholjuraev How can the feature(s) be disabled?

src/Utilities/Helpers.js
```diff
@@ -563,3 +564,29 @@ export const systemsColumnsMerger = defaultColumns => {
...
+    let environment = {};
+
+    environments.forEach(env => {
+        if (window.location.href.includes(env.key)) {
+            environment.name = env.name;
+        }
+    });
+
+    const enabledFeatures = featureFlags.filter(flag => {
+        return location.search.includes(flag.urlKey) || environment.name === 'dev' || 'stage';
```
johnsonm325, line 578: I have a few suggestions, but let me know if I'm misunderstanding how the feature flag works.

I think you mean for this line to read: `return location.search.includes(flag.urlKey) || environment.name === 'dev' || environment.name === 'stage';` In my testing, this line of code would return '`stage` if the first 2 were false, even though I was in prod.

Also, I noticed that if I added `patch_set=enabled` to the url like so: `https://stage.foo.redhat.com:1337/insights/patch/systems?page=1&perPage=20&sort=-last_upload&filter[stale]=in%3Atrue%2Cfalse&patch_set=enabled`, it would eventually reset the url to not include it. At first it would take in the parameter, but something in the code is resetting the location and then it sets it to false.

src/Utilities/constants.js
```diff
@@ -372,3 +372,25 @@ export const exportNotifications = (format) => ({
...
+    {
+        name: 'prod',
+        key: 'console.redhat.com'
+    }
+];
+
+export const featureFlags = [
+    {
+        name: 'patchSet',
+        urlKey: 'patch_set=enabled'
```
johnsonm325, line 394: This may not be best practice, but when I saw `patch_set=enabled`, I immediately thought, "what if I set `patch_set=disabled`"? Well, that would bypass the feature flag and show the content we are trying to hide. Maybe we do a check to see if it exists or is set to enabled. If it doesn't exist in the url, or it is set to disabled, then we block it.

johnsonm325 (changes requested):
@mkholjuraev Check out my suggestions and let me know if I'm misunderstanding how this all works.

src/Utilities/Helpers.js
mkholjuraev: Yeap, I did a mistake while writing this condition :)

What I wanted to achive was what you have mentioned: 
`return location.search.includes(flag.urlKey) || environment.name === 'dev' || environment.name === 'stage';`

src/Utilities/constants.js
mkholjuraev: I will fix this also.It is definately confusion by me.  While impelementing, I did not think about disabling a feature as new features are not activated by default. They are only activated if the URL has associated urlKey or they are in stage environments.

PR #752, opened 2022-03-15
chore(Travis): add custom release scripts

mkholjuraev:
After Node.js upgrade to  version 16, Travis CI build is failing. @leSamo had already this situation and suggested writing custom release. The PR is inspired by [Vulnerability](https://github.com/RedHatInsights/vulnerability-ui/blob/master/.travis/custom_release.sh). I am not familiar with Travis, so any improvements are highly welcome.

### DIFF
--- a/.travis/custom_release.sh
+++ b/.travis/custom_release.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+set -e
+set -x
+
+# map master branch to stage-beta environment
+if [[ "${TRAVIS_BRANCH}" = "master" ]]; then
+    echo "PUSHING stage-beta"
+    rm -rf ./dist/.git
+    .travis/release.sh "stage-beta"
+fi
+
+if [[ "${TRAVIS_BRANCH}" = "stage-stable" || "${TRAVIS_BRANCH}" = "prod-beta" || "${TRAVIS_BRANCH}" = "prod-stable" ]]; then
+    echo "PUSHING ${TRAVIS_BRANCH}"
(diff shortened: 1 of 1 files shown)


leSamo:
Almost perfect, but in your script you are pushing to qa-stable env from master branch and to stage-stable env from stage-stable-branch. Both qa-stable and stage-stable map to stage-stable env and qa-beta and stage-beta map to stage-beta env as shown in this [release script](https://github.com/RedHatInsights/insights-frontend-builder-common/blob/9439bff6073515eb7f7fe8046ee1d1881c332870/src/Jenkinsfile#L16):
Screenshot from 2022-03-16 10-50-23

Which would mean you are pushing to a single env from multiple branches which is undesirable and could cause confusion.

This PR will stop support for ci-beta env, I don't believe anyone is using it anymore, but better make sure.

BTW if you'd like to see the actual scripts running on Travis which will parse this custom release script you could find it here:
https://github.com/RedHatInsights/insights-frontend-builder-common

While we're at it, it would be nice to update these resources with up-to-date branch to env mappings:
https://github.com/RedHatInsights/patchman-ui#how-it-works
https://docs.engineering.redhat.com/pages/viewpage.action?spaceKey=SPM&title=Patch+UI+release+workflow

mkholjuraev:
Do you suggest to  remove  pushing to qa-stable from master?

leSamo (approved):

mkholjuraev:
@leSamo thanks for the review!

PR #755, opened 2022-03-23
feat(PatchSet): put patch set work under feature flag

mkholjuraev:
The PR enables using Unleash feature flags provided by the platform to manage path-sets feature. Current version of the Unleash is old, thus we are using only default enablement strategy. It only enables or disables the flag for now. There is a [ticket](https://issues.redhat.com/browse/APPSRE-3449) to upgrade its version. 

In order to enable/disable the flag you would need access to [Unleash instance](https://insights-stage.unleash.devshift.net/#/features/strategies/patch.patch_set) provided by app-sre. You will need to add these lines to your app-sre profile.

- $ref: /teams/insights/roles/insights-stage-unleash-admin.yml
- $ref: /teams/insights/roles/insights-unleash-admin.yml
- $ref: /teams/insights/roles/unleash-proxy.yml

For the sake of simplicity, you can test the PR by simply hard coding the flag value in the code.

### DIFF
--- a/src/Routes.js
+++ b/src/Routes.js
@@ -4,6 +4,8 @@ import React, { Fragment, lazy, Suspense, useState } from 'react';
 import { Redirect, Route, Switch } from 'react-router-dom';
 import { fetchSystems } from './Utilities/api';
 import { useHistory } from 'react-router-dom';
+import { useFeatureFlag } from './Utilities/Hooks';
+import { featureFlags } from './Utilities/constants';
 
 const Advisories = lazy(() =>
     import(
@@ -116,6 +118,8 @@ export const Routes = (props) => {
 
     const path = props.childProps.location.pathname;
 
+    const isPatchSetEnabled = useFeatureFlag(featureFlags.patch_set);
+
     return (
         // I recommend discussing with UX some nice loading placeholder
         <Suspense fallback={Fragment}>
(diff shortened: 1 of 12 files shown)


leSamo (commented):
Thank you for help with the permissions to the unleash plugin, I was able to change the flag.

I'm not sure what should be hidden by disabling the patch_set flag. I assume it is:
- if user navigates to Patch Set page, navigate him to the landing page (I assume that from the code) but that is not working for me, navigating to Patch Set page doesn't redirect me anywhere
- remove Apply to Patch Set and Remove from Patch Set buttons in tables, but even if flag is false there are these buttons in Advisory detail page:
false

Regardless of the flag being on or off clicking on Apply to Patch Set button throws an error.
patch-set

Maybe I'm doing something wrong, but I made sure to check response from `/api/featureflags/` to be what I expect.

mkholjuraev:
@leSamo thanks for the review. There was an ask to push patch-set work ASAP from management thus, this PR is not important now. I am putting this in draft state, in case it will be needed.

src/Routes.js
```diff
@@ -158,15 +162,15 @@ export const Routes = (props) => {
...
                     exact
                     path={paths.patchSet.to}
                     component={PatchSet}
-                />
+                />}
 
                 <Route
                     render={() =>
-                        some(paths, p => p.to === path) || (
+                        (!isPatchSetEnabled || !some(paths, p => p.to === path)) && (
```
leSamo, line 173: I think
`condition ? whenTrue : whenFalse`
is clearer than
`!condition || whenFalse && whenTrue`

leSamo (approved):
- When you hit Patch set navigation, it should redirect you to landing page.
    - disabled: OK
    - enabled: OK
- Systems row actions in Systems page should hide 'Assign to patch set' action.
    - disabled: OK
    - enabled: OK
- Patch set column should get hidden on Systems page.
    - disabled: OK 
    - enabled: OK
- Assign to patch sets button on Systems table toolbar should get hidden.
    - disabled: OK 
    - enabled: OK
- No patch set work should be displayed whether the flag is enabled or disabled.
    - advisory detail page has no patch set work regardless of the flag state

PR #764, opened 2022-04-07
chore(SIDs): change SIDs filter from comma separated values to single…

mkholjuraev:
Using the shared [generateFilter](https://github.com/RedHatInsights/patchman-ui/commit/efa5df4cbaf8caac2805f417f526b2d53c36617d#diff-4966f4e1c143ce9542deb4f826c300c57035c4fc796c615b9dd65aad9c3ce09bR250) caused SIDs filter to change to comma separated multivalue filter (`filter[system_profile][sid_filter][in]=a,v`). The engine team asked to avoid this pattern due to perfomance issues. This PR brings back original single value filters pattern: `filter[system_profile][sap_sids][in]=AV1&filter[system_profile][sap_sids][in]=AV2&limit=1&offset=0`

### DIFF
--- a/src/Utilities/Helpers.js
+++ b/src/Utilities/Helpers.js
@@ -245,14 +245,16 @@ export const getFilterValue = (category, key) => {
 };
 
 export const encodeParams = (parameters, shouldTranslateKeys) => {
-    const calculateWorkloads = (systemProfile) => {
+    const calculateWorkloads = ({ sap_sids, ...restOfProfile }) => {
         let result = '';
-        Object.entries(generateFilter({ system_profile: systemProfile })).forEach(entry => {
+        Object.entries(generateFilter({ system_profile: restOfProfile })).forEach(entry => {
             const [key, value] = entry;
             result = `${result}&${key}=${value}`;
         });
 
-        return result;
+        const SIDsFilter = sap_sids?.map(sid => `filter[system_profile][sap_sids][in]=${sid}`).join('&');
+
+        return result.concat(sap_sids ? `&${SIDsFilter}#SIDs=${sap_sids.join(',') }` : '');
     };
 
     const flattenFilters = filter => {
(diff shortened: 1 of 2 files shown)


johnsonm325 (changes requested):
I'm only seeing one problem, and I'm not sure what the expected action is. The url is correctly set on global filter update, but if you set the global filters in the url manually, it does not update the filters in the UI.

For example, if I have no filters set in the UI, and then I paste the following in to the url, `https://stage.foo.redhat.com:1337/insights/patch/advisories?page=1&page_size=20&offset=0&sort=-applicable_systems&filter[system_profile][sap_sids][in]=AV1#SIDs=&tags=` and press enter, the global filters will not be set in the UI.

Drift actually does this correctly, though I don't think drift follows the correct url pattern. The problem is, I can't remember how it was done. I'm either going to look more tonight or take another look on Monday. Here's where it _might_ be happening: https://github.com/RedHatInsights/drift-frontend/blob/master/src/App.js#L86-L92

Let me know if you think we should fix this in another PR and I'll approve this one.

mkholjuraev:
@johnsonm325 thanks for the review. I am working on a fix.

mkholjuraev:
@johnsonm325 I have found out why this is happening. It is because patch workloads filter syntax is different from Global filter syntax in the URL. This is causing Global filter no to trigger when you enter `filter[system_profile][sap_sids][in]=AV1` manually. If I add `&SIDs=AC1#SIDs=AC1&tags=` in the URL Global filter gets triggered. 

The fix can be adding this part of url from global filter into the url not from Patch SIDs filter. But this gets hackish :(. I need to find the optimal solution...

mkholjuraev:
Ups, I will fix just a sec

johnsonm325:
@mkholjuraev if you add `&filter[system_profile][sap_system]=true` in the url and load it, it does not add the SAP workload filter. I think you'll need to add `&workloads=`

mkholjuraev:
@johnsonm325 If I understand correctly, this is expected, it is because Global filter does not react to this filter syntax. You need to add #sids=some_sid instead. As I have mentioned the filter syntax is different on patch engine side and Global filter from the platform side. I am not sure how I can force the imported global filter react to `&filter[system_profile][sap_system]=true` when it is manually entered in the URL. 

I can write some js hackish code, but I believe that it has more disadvantages than advantages. Please let me know your feedback.

johnsonm325:
These workloads confuse me. I need an education in how they work. If I hit the following url in drift, `https://console.stage.redhat.com/insights/drift#workloads=SAP&SIDs=&tags=`, the SAP workload is set and when I load systems, I only get SAP systems.

mkholjuraev:
Lets have a short call tomorrow to discuss our findinds if you do not mind.

mkholjuraev:
@johnsonm325 as we have discussed, I am merging this PR. I have created a new ticket for the global filter issue: https://issues.redhat.com/browse/SPM-1446

mkholjuraev:
Thanks for the review!

PR #789, opened 2022-04-26
chore(InventoryDetail) remove feature flag, show patch set only when it is present

mkholjuraev:
I had discussion with UX and they agreed to remove completely patch set info on System detail header when it is not present.

Also this PR removes patch set flag **Assign to patch set** on system detail actions.


gabipodolnikova (approved):
Yes, this is better than having a blank space. Thanks! :)

PR #790, opened 2022-04-26
chore(ReviewSystems): fix filters breaking the page

mkholjuraev:
This resolves: https://issues.redhat.com/browse/SPM-1461.

To repoduce the issue: 

1. Open the patch set wizard
2. Proceed to Review Systems step
3. Active multiple filter with a light speed
4. Hit on the 'Reser filters' button. At this step, the wizard should break and close.


mkholjuraev:
@RedHatInsights/team-interact I need to make a new release before code freeze. Thus I am merging this PR (as it is only a small fix) if you do not mind.

PR #796, opened 2022-05-12
fix(1494): fix selected systems across steps

mkholjuraev:
This resolves the ticket: https://issues.redhat.com/browse/SPM-1494

To reproduce:

1. Create new patch set using wizard
2. On 2nd page select e.g. 3 systems, deselect 2 of them so 1 is selected.
3. Go to Final page of the wizard, the wizard shows 1 system in patch set
4. Go back to system selection, 3 systems are selected although no action was taken.
5. Go next to final page of the wizard, it shows 3 systems now although only actions taken were going to previous (2nd page) and immediately to 3rd page.


adonispuente (approved):
I think this fix works great, I don't see any issues!

mkholjuraev:
@adonispuente sounds great!

PR #797, opened 2022-05-12
fix(SPM-1492): system details page refreshs after patch-set change

mkholjuraev:
Resolves: https://issues.redhat.com/browse/SPM-1492. I am not able to do API request other than the type GET, theoratically just adding `patchSetState.shouldRefresh` to the the hook dependency should do the trick, most probably. Thus, please run this PR locally to make sure it does the fix.

To reproduce:
1. Assign system to a patch set via action on system detail page
2. The page continues to display old patch set name a needs to be refreshed to display new patch set name

### DIFF
--- a/src/SmartComponents/SystemDetail/InventoryDetail.js
+++ b/src/SmartComponents/SystemDetail/InventoryDetail.js
@@ -37,7 +37,7 @@ const InventoryDetail = ({ match }) => {
         return () => {
             dispatch(clearNotifications());
         };
-    }, []);
+    }, [patchSetState.shouldRefresh]);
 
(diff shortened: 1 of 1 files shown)


MichalSajdik (commented):
I noticed that in https://github.com/RedHatInsights/patchman-ui/pull/798
you have similiar changes with different commit id's. I think it should be consistent

MichalSajdik (changes requested):
the comment above

mkholjuraev:
@MichalSajdik thanks for having a look. Yeap, I forgot to checkout to a new branch from master branch. Issue is resolved.

mkholjuraev:
@MichalSajdik thanks!

PR #804, opened 2022-05-19
fix(SPM-1488): remove systems from patch sets

mkholjuraev:
Resolves: https://issues.redhat.com/browse/SPM-1488. I am not able to make any other API requests except GET. Thus please make sure that API handles request nicely.

There are 2 use cases where users can remove system(s) from patch sets.

To reproduce from table row actions:
1. Go to Systems page
2. Click on 'Remove from patch set' row action of a system which has a set assigned.
3. The confirmation with the system number should appear
4. When users clicks on Remove button on the modal, system should be unassigned from the patch set

 To reproduce from table toolbar dropdown:
 1. Go to Systems page
 2. Select as many systems as you want
3. Click on 'Remove from patch set' button from the tabletoolbars dropdown.
4. The confirmation with the system number should appear
5. When users clicks on Remove button on the modal, system should be unassigned from the patch set

### DIFF
--- a/src/Messages.js
+++ b/src/Messages.js
@@ -41,6 +41,11 @@ export default defineMessages({
         description: 'button label',
         defaultMessage: 'Edit patch set'
     },
+    labelsCancel: {
+        id: 'labelsCancel',
+        description: 'Button label',
+        defaultMessage: 'Cancel'
+    },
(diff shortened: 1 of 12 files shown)


mkholjuraev:
@MichalSajdik :+1:

PR #805, opened 2022-05-26
fix(PatchSetWizard): Adding missing message

gabipodolnikova:
I am not sure what the title should be, but there was a missing message causing a blank page after assigning a patch set.

### DIFF
--- a/src/Messages.js
+++ b/src/Messages.js
@@ -394,7 +394,7 @@ export default defineMessages({
     patchSetTitle: {
         id: 'patchSetTitle',
         description: 'title of the patch set wizard',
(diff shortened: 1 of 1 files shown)


mkholjuraev (commented):
@gabipodolnikova thanks for the help. This messages should have somehow disappeared while rebasing commits. Can I please ask to change the message from 'Patch set' to 'Create Patch set'.

mkholjuraev (approved):
@gabipodolnikova  :+1: . I could not run this locally due to VPN, but I am sure that this is the fix.

mkholjuraev:
Merging if you do not mind.

PR #807, opened 2022-05-31
feat(SPM-1311): enable patch set removal on System details page

mkholjuraev:
Enables system removal from a patch set on system details page (https://issues.redhat.com/browse/SPM-1311).

To verify, visit a system details page

1.  when that system has a set assigned, you should see 'Remove from patch set' action enabled on the right corner dropdown.

- the action should open a modal with confirmation purpose
- system should get unassigned, when user  hits on 'Remove' button on the modal.
- page should refresh after successful removal

2. when that system has no set assigned, you should see  'Remove from patch set' action disabled on the right corner dropdown

### DIFF
--- a/src/SmartComponents/Modals/UnassignSystemsModal.js
+++ b/src/SmartComponents/Modals/UnassignSystemsModal.js
@@ -14,19 +14,20 @@ const UnassignSystemsModal = ({ unassignSystemsModalState = {}, setUnassignSyste
 
     const { systemsIDs, isUnassignSystemsModalOpen } = unassignSystemsModalState;
 
-    const handleModalOpen = () => {
+    const handleModalOpen = (shouldRefresh) => {
         setUnassignSystemsModalOpen({
             isUnassignSystemsModalOpen: !isUnassignSystemsModalOpen,
-            systemsIDs: []
+            systemsIDs: [],
+            shouldRefresh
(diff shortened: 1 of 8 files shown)


mkholjuraev:
@RedHatInsights/team-interact I would appreciate a look on this PR.

gabipodolnikova (approved):
Looking good! 👍

src/SmartComponents/Modals/UnassignSystemsModal.js
```diff
@@ -14,19 +14,20 @@ const UnassignSystemsModal = ({ unassignSystemsModalState = {}, setUnassignSyste
 
     const { systemsIDs, isUnassignSystemsModalOpen } = unassignSystemsModalState;
 
-    const handleModalOpen = () => {
+    const handleModalOpen = (shouldRefresh) => {
         setUnassignSystemsModalOpen({
             isUnassignSystemsModalOpen: !isUnassignSystemsModalOpen,
```
gkarat, line 19: just a minor thing: if `handleModalOpen` is called only when there is confirm or close button clicked, then it is better to pass `false` explicitly to the `isUnassignSystemsModalOpen` flag:

```suggestion
            isUnassignSystemsModalOpen: false,
```

PR #813, opened 2022-06-03
fix(SPM-1457): enable up-to-date systems to be operated on

mkholjuraev:
Resolves  [SPM-1457](https://issues.redhat.com/browse/SPM-1457) by enabling up-to-date systems to be bulk selected. NoDataModal is displayed when an up-to-date system is chosen to be remediated. In case, more than one of those systems are passed to Remediation with remediatable systems, that modal does not show up.This is consistent with current Remediation application.

This PR also disables appliying remediation to up-to-date systems on table row actions.

To test:

1.  Navigate to Systems page.
2. Select any one system that is up-to-date. ( you can decide if a system is up-to-date by looking at the table row actions. If 'Apply all applicable advisories' action is disabled, it means that this  system is up-to-date).
3. Click Remediate button on the table toolbar.
4. NoDataModal should show up.


src/SmartComponents/Systems/Systems.js
```diff
@@ -169,12 +164,18 @@ const Systems = () => {
...
     const getEntities = useGetEntities(fetchSystems, apply, {}, history, applyMetadata, applyGlobalFilter);
 
-    const remediationDataProvider = () => remediationProviderWithPairs(
-        removeUndefinedObjectKeys(selectedRows),
-        prepareRemediationPairs,
-        transformPairs,
-        remediationIdentifiers.advisory
-    );
+    const remediationDataProvider = () => {
+        setRemediationLoading(true);
```
gabipodolnikova, line 168: If we are setting the remedationLoading here, is it needed in `showRemediationModal`?

mkholjuraev: @gabipodolnikova thank you for the review, actually it is present there as well. Git did not show that line becuase it was present before this PR and Git did not detect any change.

gabipodolnikova (approved):
LGTM! 😎

PR #814, opened 2022-06-03
Batch systems removal

mkholjuraev:
This PR is to resolve https://issues.redhat.com/browse/SPM-1488 by enabling users remove one or more systems from one or more patch sets. The action can be taken through the 'Remove from Patch Set' action in the table toolbas and the table row actions.

Expectations:

1. When you try to remove a system without patch set assigned, 'Remove from Patch Set' row action should be disabled.
2. When you try to remove a single system without patch set assigned, a modal should pop with proper message indicating there is no effect and disabled Remove button.
3. When you select a single system with a patch set assigned, modal should open and you should be able to remove.
4. If you select systems with both patch set assigned and not assigned, a modal with the message about  _Do you want to remove the SomeNumber selected system from assigned Patch sets?_ and _There is 1 system you are trying to remove that is not assigned to any existing Patch Set. This action will not affect it._


src/SmartComponents/Systems/Systems.js
```diff
@@ -218,6 +219,7 @@ const Systems = () => {
                                     selectedTags
                                 },
                                 shouldRefresh: patchSetState.shouldRefresh === true
+                                    || unassignSystemsModalState.shouldRefresh === true
```
gabipodolnikova, line 222: Wouldn't just `shouldRefresh: patchSetState.shouldRefresh || unassignSystemsModalState.shouldRefresh` be enough?

gabipodolnikova (changes requested):
Just two minor things.
And also leaving what I already sent you here too:
When removing patch set from the toolbar it does not refresh the page.
When assigning either from row actions or a toolbar it also does not refresh the page.

mkholjuraev:
@gabipodolnikova thanks for the review. All suggestions are improvement and I have applied them. The refresh bug is also fixed hopefully.

src/SmartComponents/Modals/UnassignSystemsModal.js
```diff
@@ -23,41 +25,69 @@ const UnassignSystemsModal = ({ unassignSystemsModalState = {}, setUnassignSyste
...
         //TODO: mockups do not have error notifications designed, add them if UX designes.
         if (result.status === 200) {
             handleModalOpen(true);
-            dispatch(addNotification(patchSetUnassignSystemsNotifications(systemsIDs?.length || 0).success));
+            dispatch(addNotification(patchSetUnassignSystemsNotifications(systemsWithPatchSet?.length || 0).success));
         }
     };
 
+    useEffect(() => {
+        //filters out systems without patch sets assigned
```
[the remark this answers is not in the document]
mkholjuraev: Moved into helper function :)

src/SmartComponents/Systems/Systems.js
mkholjuraev: I have used onRefreshData function to relaod Inventory component. This helped to make the code to be readable, I think.

gabipodolnikova (approved):
Looks good, thank you! :)

PR #820, opened 2022-06-15
fix(api): errors from api response are passed down to components

mkholjuraev:
The PR is intended to passdown api error response to the smart components. An error response is used to detect error status and handle correspondingly inside Smart components. e.g ([Systems.js](https://github.com/RedHatInsights/patchman-ui/blob/master/src/SmartComponents/Systems/Systems.js#L212) receives error status code and passes that to [ErrorHandler](https://github.com/RedHatInsights/patchman-ui/blob/master/src/PresentationalComponents/Snippets/ErrorHandler.js) ). Furthermore, redundant user checking before an api call is removed.

It also refactors test coverage code for api.js:

- disabled eslint rules are removed
- first test case for JSON response is removed as it is not necessary anymore
- some dirty code cleanup.

### DIFF
--- a/src/Utilities/api.js
+++ b/src/Utilities/api.js
@@ -1,5 +1,3 @@
-/* eslint-disable camelcase */
-/* eslint-disable no-unused-vars */
 import axios from './axiosInterceptors';
 import { encodeApiParams, prepareEntitiesParams } from './Helpers';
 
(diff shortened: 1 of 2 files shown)


AsToNlele (approved):
Looks good!

mkholjuraev:
@AsToNlele thank you for your quick response!

PR #821, opened 2022-06-15
fix(TableView): wrap table toolbar with ErrorHandler

mkholjuraev:
This will hide the table toolbar when there is an some error.

Before: 

image

After:
image


mkholjuraev:
@AsToNlele thanks for the review. I will apply your suggestion and then merge!

PR #823, opened 2022-06-18
fix(InventoryDetail): ESSNTL-2698 handle disconnected systems

mkholjuraev:
Resolves: [ESSNTL-2698](https://issues.redhat.com/browse/ESSNTL-2698).

To reproduce:

1. run this app against stage env
2. visit patch systems page
3. click on any system that is disconnected from insights-client (this kind of systems are marked with disconnected icon in the last seen column).
4. System details page should have special empty component for disonnected systems.


mkholjuraev (commented):
I tried reviewing the PR. Looks good to me :).

PR #834, opened 2022-06-29
Rename patch set

mkholjuraev:
Renames all occurances of 'Patch set' on the UI to 'Patch template'. There are some places only 'Template' word is used. As the mockups are not updated in detail, I tried to cover all places and think about correct words, where to rename... 

Only wording on the UI should change without effecting functionality.

Please refer to the [mockups](https://marvelapp.com/prototype/hddg13b/screen/86218222) for more information.


src/SmartComponents/PatchSet/PatchSetAssets.js
```diff
@@ -37,7 +37,7 @@ export const patchSetRowActions = (openPatchSetEditModal, handlePatchSetDelete)
         }
```
gabipodolnikova, line 37: And 'Edit patch set' is staying in row actions

gabipodolnikova (commented):
Found one possible missing rename and also the navigation item Patch set should be renamed in the config repo right? :)

src/SmartComponents/PatchSet/PatchSetAssets.js
mkholjuraev: @gabipodolnikova thank you for the quick look. Nope, I will rename this line as well :)

mkholjuraev: @gabipodolnikova I have fixed ^. You can find it easly on a new commit.

gabipodolnikova (approved):
LGTM! Thanks! :)

PR #843, opened 2022-07-02
chore(PatchTemplate): internationalize wizard, optimize text rendering

mkholjuraev:
Internationalizes texts on Patch template with text render optimization.

### DIFF
--- a/src/Messages.js
+++ b/src/Messages.js
@@ -41,6 +41,11 @@ export default defineMessages({
         description: 'button label',
         defaultMessage: 'Edit template'
     },
+    labelsButtonRemoveTemplate: {
+        id: 'labelsButtonRemoveTemplate',
+        description: 'button label',
+        defaultMessage: 'Remove template'
+    },
     labelsCancel: {
         id: 'labelsCancel',
         description: 'Button label',
@@ -135,6 +140,11 @@ export default defineMessages({
         id: 'labelsCvesButton',
         description: 'label for cves button',
         defaultMessage: '{cvesCount, plural, one {# CVE} other {# CVEs}} associated with this patch'
+    },
+    labelsDate: {
+        id: 'labelsDate',
+        description: 'Label',
+        defaultMessage: 'Date'
     },
     labelsDescription: {
         id: 'labelsDescription',
@@ -336,6 +346,11 @@ export default defineMessages({
         description: 'Return to previous page label for general usage',
         defaultMessage: 'Return to previous page'
     },
+    labelsSelectedSystems: {
+        id: 'labelsSelectedSystems',
+        description: 'label',
+        defaultMessage: 'Selected systems'
+    },
     labelsSeverity: {
         id: 'labelsSeverity',
         description: 'label for cves info',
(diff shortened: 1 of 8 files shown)


src/Messages.js
```diff
@@ -496,11 +536,31 @@ export default defineMessages({
         description: 'text to notify there is not available version',
         defaultMessage: 'No version is available'
     },
+    textPatchTemplatePending: {
+        id: 'textPatchTemplatePending',
+        description: 'text for the patch template',
+        defaultMessage: 'Please allow a few minutes to set up a patch template. \n You will receive a notification when finished'
```
gkarat, line 542: a very nit-pick:

```suggestion
        defaultMessage: 'Please allow a few minutes to set up a patch template. \n You will receive a notification when finished.'
```

src/SmartComponents/PatchSetWizard/steps/RequestProgress.js
```diff
@@ -44,41 +46,44 @@ const RequestProgress = ({ onClose, state }) => {
...
             </Title>
             <EmptyStateBody>
                 <Grid hasGutter>
                     {(requestPending) && (
                         <><GridItem>
                             <HelperText>
                                 <HelperTextItem variant="indeterminate">
-                                    Please allow a few minutes to set up a patch template.
-                                    You will receive a notification when finished</HelperTextItem>
+                                    {intl.formatMessage(messages.textPatchTemplatePending)}
```
gkarat, line 60: I didn't run it locally, but did you check whether the newline was rendered here? IIRC, I had a problem with newliners while formatting messages with formatjs

src/SmartComponents/PatchSetWizard/steps/ReviewPatchSet.js
```diff
@@ -11,6 +11,19 @@ import {
     TextListItemVariants
 } from '@patternfly/react-core';
 import useFormApi from '@data-driven-forms/react-form-renderer/use-form-api';
+import { intl } from '../../../Utilities/IntlProvider';
+import messages from '../../../Messages';
+
+const renderTextListItem = (label, text) => (
+    <Fragment>
```
gkarat, line 18:
```suggestion
    <React.Fragment>
```
because the rest of the app uses this notation :) better to choose one and be consistent here..

src/SmartComponents/PatchSetWizard/steps/ReviewPatchSet.js
```diff
@@ -11,6 +11,19 @@ import {
...
+        <TextListItem component={TextListItemVariants.dt}>
+            {intl.formatMessage(label)}:
```
gkarat, line 20: Maybe it is more clear to pass `label` as a string formatted in advance? I mean, `label` is a message descriptor, whereas `text` is not (it is already a string and was pre-formatted before passing it to this function).

gkarat (approved):
Just minor things, otherwise LGTM

src/Messages.js
mkholjuraev: Nice catch!

src/SmartComponents/PatchSetWizard/steps/RequestProgress.js
mkholjuraev: The actual expectation is to make it inline :)

mkholjuraev:
@gkarat thank you for the review!

PR #844, opened 2022-07-07
fix(Date): improve validation, introduce date start limit

mkholjuraev:
This is to improve date field validation and introduce start limit.

To check locally:

- make sure that the date input does not let users select a date prior to 1990.01.01
- make sure that input field validates: wrong date, date input format


adonispuente (approved):
This change works great, LGTM!
I didn't know the comparison operators '>' would work with dates like that

mkholjuraev:
@adonispuente thank you!

PR #849, opened 2022-07-22
chore(Unleash): disable unleash usage and use isBeta flag

mkholjuraev:
Use the isBeta flag from chrome instead of unleash until unleash becomes mature.


bastilian (approved):
LGTM! Thank you @mkholjuraev!

mkholjuraev:
@bastilian thank you for the review!

PR #857, opened 2022-07-29
fix(select): SPM-1418 Fix selection on PackageSystems

jiridostal:
With this commit, remediation should work as expected.


mkholjuraev (approved):
LGTM! thanks @jiridostal !

PR #858, opened 2022-08-01
feat(RHIF-83: introduce loading icon on remediation button in Advisor…

mkholjuraev:
Partly resloves: https://issues.redhat.com/browse/RHIF-51.

It just introduces a loading icon while the remediation issues map is being built. There is a new ticket in the advisor to optimize the API requests to fetch the issues map.

### DIFF
--- a/src/PresentationalComponents/TableView/TableView.js
+++ b/src/PresentationalComponents/TableView/TableView.js
@@ -39,7 +39,8 @@ const TableView = ({
     defaultFilters,
     searchChipLabel,
     CreatePatchSetButton,
-    actionsConfig
+    actionsConfig,
+    isRemediationLoading
 }) => {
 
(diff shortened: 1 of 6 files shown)


adonispuente (approved):
Thank you, PR works as intended, nice organization. LGTM!

Just a side note, has it been considered to disable the remediate button if the selected items are remediable automatically by ansible? I know theres a popup that displays afterwards if all remediations are manual, but it takes a bit to get feedback from it. There could be a reason we dont do this, but I just wanted to through it out there and hear your thoughts!

mkholjuraev:
@adonispuente thank you for the review! YOU are awesome!

mkholjuraev:
@adonispuente in a good world we should not see that modal at all on Advisories page. That modal is poping up because of API request failure. I have asked the backend to fix that endpoint.

PR #861, opened 2022-08-04
fix(BulkSelect): enable loading state when all entities selected

mkholjuraev:
This introduces a loading icon when all entities are selected using the bulk selector.

### DIFF
--- a/src/PresentationalComponents/TableView/TableView.test.js
+++ b/src/PresentationalComponents/TableView/TableView.test.js
@@ -95,7 +95,7 @@ describe('TableView', () => {
         const wrapper = shallow(<TableView {...testObj} />);
         const { bulkSelect: { items }}  = wrapper.find('PrimaryToolbar').props();
         items[2].onClick();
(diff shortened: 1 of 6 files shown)


AsToNlele:
Thank you for your PR!
I've found in some really rare occasions there are two labels visible, but it would show up randomly and very few times
image
Looks great otherwise!

mkholjuraev:
@AsToNlele great catch! Can you tell in what cases, does this occur?

AsToNlele:
I'm afraid its completely random, It happened to me on the first try and then after like 10 tries.

mkholjuraev:
@AsToNlele the issue is fixed

adonispuente:
The ' Select All’ option seems to be working perfectly fine, but the loading icon doesnt pop up when just clicking the empty square, even though clicking the empty square does the same action. I believe we should add this icon when clicking this button

mkholjuraev:
@adonispuente thanks for the suggestion. I have enabled the loader for that action as well. Please have a look.

adonispuente (approved):
great PR, LGTM!

PR #870, opened 2022-09-08
chore(cleanup): remove location listening hook and some cleanup

mkholjuraev:


mkholjuraev (commented):
Merging as it is just a small cleanup

PR #871, opened 2022-09-09
chore(bulkSelect): move useBulkSelect invocation out of jsx sections

mkholjuraev:


mkholjuraev (commented):
Merging as there is not a change in the codebase, only order of function call is changed

PR #882, opened 2022-10-06
Fix(InventoryReloadBug): changes how to store and registry are set up 

adonispuente:
This is a potential fix for the inventory reloading bug.

This version of the PR just includes change the registry/store in AppEntry. With no Context Provider component

### DIFF
--- a/src/AppEntry.js
+++ b/src/AppEntry.js
@@ -1,21 +1,39 @@
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import { Provider } from 'react-redux';
 import { BrowserRouter as Router } from 'react-router-dom';
 import App from './App';
-import { getStore } from './store';
+// import { getStore } from './store';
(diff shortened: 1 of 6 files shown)


src/AppEntry.js
```diff
@@ -1,21 +1,36 @@
...
 import { getStore } from './store';
 import getBaseName from './Utilities/getBaseName';
-import { IntlProvider } from '@redhat-cloud-services/frontend-components-translations';
-import messages from '../locales/en.json';
+import { init, RegistryContext } from './store';
+import logger from 'redux-logger';
+import { EmptyState } from '@patternfly/react-core';
 
-const Patchman = () => (<div className="patch-root">
-    <IntlProvider locale={navigator.language.slice(0, 2)} messages={messages}>
```
bastilian, line 11: This should probably not be removed.

adonispuente: Ill slap it back in now

PR #891, opened 2022-10-19
improve paginated remediation pairs fetching

mkholjuraev:
This PR will improve paginated Remediation issues fetching by 

1. creating grouped API requests with time interval
2. when all items are selected, only /view/advisory/systems endpoint is used. This endpoint behaves the same when all entities are requested and is faster.
3. limit and offset are introduced into the API requests to speed up the SQL query.

Please verify that there is a correct number of API requests to remediate all selected entities and that pagination works as it should.


src/Utilities/Hooks.js
```diff
@@ -14,7 +16,8 @@ import { multiValueFilters } from '../Utilities/constants';
 import { assignSystemPatchSet, updatePatchSets } from './api';
 // eslint-disable-next-line no-unused-vars
 import { useFlag, useFlagsStatus } from '@unleash/proxy-client-react';
```
adonispuente, line 18: This doesnt seem to be used

adonispuente (commented):
While testing it seems to work as intended. However I noticed that going through the actually remediation process is very slow. When I try to complete remediation, it errors out every time

mkholjuraev:
@adonispuente thank you for the review. I have already started refactoring and preparing remediation issues in the patch for a web worker. This means that all of this work will be done outside of the main thread. Surely, this will optimize the process, however, the remediation app still takes too much time and we can not do something with this from the Patch side.

adonispuente (approved):
PR works as intended, a ticket needs to be created for the remediation side of things as discussed on call. LGTM!

mkholjuraev:
@adonispuente thank you.

PR #892, opened 2022-10-25
Webwork

mkholjuraev:
Resolves: # RHIF-106
This a separate PR to improve paginated Remediation issues fetching by

1. Moving all cumbersome jobs into a web worker
2. creating grouped API requests with time interval
3. limit and offset are introduced into the API requests to speed up the SQL query.
4. disables remediation button, after it is already clicked and the module is being loading 

Please verify that there is a correct number of API requests to remediate all selected entities and that pagination works as it should.


mkholjuraev:
@RedHatInsights/team-interact I am ready to do a pair reviewing if this helps ;).

src/Utilities/RemediationPairs.js
```diff
@@ -0,0 +1,131 @@
...
+                await dataFetcher({
+                    ...payloadModifier(payload),
+                    limit: REQUEST_CHUNK_SIZE,
+                    offset: ((batchIndex * BATCH_REQUEST_SIZE) + i) * REQUEST_CHUNK_SIZE
+                }
+                ).then(result => responseTransformer(result))
+            );
+        }
+
+        return await Promise.all(requests);;
```
bastilian, line 79: I don't think, you need the `await` or the `Promise.all` here. The Promises pushed into the `request` array are already `await`ed for when looping over the groupChunks and pushed into the array.

mkholjuraev: I thought this will ensure all promises to finish their call. I will verify if they are indeed not necessary and update my PR if needed.

mkholjuraev: @bastilian one more benefit that was in my mind was if I wrap all those API calls promises into a single Promise.all, I would ensure that no remaining promises resolve if any one fails. Considering this, I would keep this. (Feel free to suggest a better way if you have)

bastilian: I'm not 100% certain that is how it would behave. If any of the promises would fail/error and can't be awaited for on line 70, wouldn't they raise and error up there before even reaching the `Promise.all`?

mkholjuraev: hmm, yes it makes sense. I will remove this

mkholjuraev: Done.

src/Utilities/useRemediationDataProvider.test.js
```diff
@@ -1,116 +1,133 @@
...
-
-jest.mock('./api', () => ({
-    ...jest.requireActual('./api'),
-    fetchViewAdvisoriesSystems: jest.fn().mockReturnValue(new Promise((resolve) => {
-        resolve({ data: { testAdvisory: ['test-system'] } });
-    })),
-    fetchViewSystemsAdvisories: jest.fn().mockReturnValue(new Promise((resolve) => {
-        resolve({ data: { testAdvisory: ['test-system'] } });
-    }))
+describe('useRemediationDataProvider', () => it('temproray test', () => {
```
bastilian, line 1: Maybe we can use the react testing library to add tests for this hook? https://react-hooks-testing-library.com

bastilian: There are examples of it being used in compliance to test hooks: https://github.com/RedHatInsights/compliance-frontend/blob/master/src/Utilities/hooks/useTableTools/useActionResolver.test.js

mkholjuraev: The issue I faced was more related to mocking the import of a webworker using `import.meta.URL`. There is a library that enables mocking it by configuring babel, but this was causing the webwork not to load into the hook. I am still looking for a way to properly test this.

src/Utilities/RemediationPairs.js
```diff
@@ -0,0 +1,131 @@
...
+                    pairs[index].systems.push(systemID);
+                } else {
+                    pairs.push(
+                        {
+                            id: pairID,
+                            description: advisory,
+                            systems: [systemID]
+                        }
+                    );
+                }
```
gkarat, line 52: A nit-pick: this seems to be inefficient, especially when you search for indices with `findIndex`. Probably, you could use `reduce` here together with `pairs` set as an object, not array? This will let you easier work and update the dictionary (object). In the end, you can transform it back to array...

gkarat (approved):
lgtm! tested with the steps you provided, works ok for accounts with lesser number of advisories/systems. a few comments on the code: I think there are places where you can work with data more efficiently by reducing loops

src/Utilities/RemediationPairs.js
mkholjuraev: Actually, I started pairs as an object initially but later had to convert it to array type. Let me remember why I switched to Array.

PR #895, opened 2022-11-01
Support component tests

gkarat:
This PR installs cypress-related dependencies and makes sure we can run component tests. The cypress bundle is now also included to the Test stage in Travis CI (this doesn't modify pr_check since we are not changing `verify` npm script).

изображение


mkholjuraev (approved):
LGTM!

PR #906, opened 2022-11-04
(SPM-1752): RBAC implementation | Templates Page permissions

adonispuente:
In accordance with https://issues.redhat.com/browse/SPM-1768
This adds RBAC permissions to patch. If navigating to Templates page with only patch viewer permissions, a user cannot select the 'Create a Template' button and a tooltip hovers over the button. The same goes for editing a template.


src/PresentationalComponents/WithPermission/WithPermission.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import propTypes from 'prop-types';
+import { usePermissions } from '@redhat-cloud-services/frontend-components-utilities/RBACHook';
+import { NotAuthorized } from '@redhat-cloud-services/frontend-components/NotAuthorized';
+
+const WithPermission = ({ children, requiredPermissions = [] }) => {
+    const { hasAccess, isLoading } = usePermissions('patch', requiredPermissions);
```
bastilian, line 7: `usePermissions` can be a bit wonky at times, i think using the `usePermissionWithContext` with the RBACProvider setup in the App/Entry.js(somewhere above) would be better.

src/Routes.js
```diff
@@ -46,102 +73,78 @@ const Templates = lazy(() =>
...
+        requiredPermissions: ['patch:*:read'],
+        component: Templates
     }
-};
-
-export const Routes = (props) => {
-
-    const path = props.childProps.location.pathname;
-
-    const isPatchSetEnabled = useFeatureFlag(featureFlags.patch_set);
```
mkholjuraev, line 92: Are we safe to remove the flag already? if not it is better to leave it for now.

adonispuente: Ahh this was a mistake, ill add this back

src/SmartComponents/AdvisoryDetail/AdvisoryDetail.js
```diff
@@ -51,7 +51,7 @@ const AdvisoryDetail = ({ match }) => {
                 breadcrumbs={[
                     {
                         title: intl.formatMessage(messages.titlesPatchAdvisories),
-                        to: paths.advisories.to,
```
bastilian, line 54: It could still be preserved that the paths come from `paths`, but it would need to use `path` instead of `to`.

src/Routes.js
```diff
@@ -46,101 +82,94 @@ const Templates = lazy(() =>
...
+                    <Spinner />
+                </Bullseye>
+            }
+        >
             <Switch>
+                {paths.map(PermissionRouter)}
                 <Redirect
-                    from={paths.advisoryDetailSystem.to}
-                    to={paths.inventoryDetail.to}
+                    from='/advisories/:advisoryId/:inventoryId'
```
bastilian, line 159: Here we should probably also preserve these being grabbed from `paths`.

src/SmartComponents/AdvisoryDetail/__snapshots__/AdvisoryDetail.test.js.snap
```diff
@@ -81,7 +81,7 @@ exports[`AdvisoryDetail.js Should match the snapshots 1`] = `
                 Object {
                   "isActive": false,
                   "title": "Patch advisories",
-                  "to": "/advisories",
+                  "to": undefined,
```
bastilian, line 84: Maybe the test-data/fixtures needs to be adjusted here?

src/SmartComponents/PatchSet/PatchSet.js
```diff
@@ -114,9 +115,12 @@ const PatchSet = ({ history }) => {
...
     };
 
-    const CreatePatchSetButton = createPatchSetButton(setPatchSetState);
+    const { hasAccess } = usePermissionsWithContext([
+        'patch:*:*'
+    ]);
+    const CreatePatchSetButton = createPatchSetButton(setPatchSetState, hasAccess);
     const actionsConfig = patchSetRowActions(openPatchSetEditModal, handlePatchSetDelete);
-
+    const customActionsConfig = customActionsToggle;
```
bastilian, line 123: This seems a bit redundant, we could just use `customActionsToggle` below on line 158.

src/SmartComponents/PatchSet/PatchSetAssets.js
```diff
@@ -42,4 +51,15 @@ export const patchSetRowActions = (openPatchSetEditModal, handlePatchSetDelete)
             handlePatchSetDelete(rowData);
         }
     }
+
 ];
+
+export const customActionsToggle = () => <Tooltip content='For editing access, contact your administrator.'>
```
bastilian, line 57: `CustomActionsToggle`

src/SmartComponents/SystemDetail/InventoryDetail.js
```diff
@@ -11,14 +10,17 @@ import { setPageTitle, useFeatureFlag } from '../../Utilities/Hooks';
...
 import { clearNotifications } from '@redhat-cloud-services/frontend-components-notifications/redux';;
 import ErrorHandler from '../../PresentationalComponents/Snippets/ErrorHandler';
 import PatchSetWrapper from '../../PresentationalComponents/PatchSetWrapper/PatchSetWrapper';
 import usePatchSetState from '../../Utilities/usePatchSetState';
 import { featureFlags } from '../../Utilities/constants';
+import { paths } from '../../Routes';
+import { useParams } from 'react-router-dom';
+
+const InventoryDetail = () => {
+    const {  inventoryId } = useParams();
```
bastilian, line 22: I'm not sure why yet, but this triggers a 🚩in me.

src/SmartComponents/SystemDetail/InventoryDetail.js
```diff
@@ -27,7 +29,7 @@ const InventoryDetail = ({ match }) => {
     const { display_name: displayName, insights_id: insightsID } = useSelector(
         ({ entityDetails }) => entityDetails?.entity ?? {}
     );
-    const entityId = match.params?.inventoryId;
+    const entityId = inventoryId;
```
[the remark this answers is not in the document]
adonispuente: Woops

src/SmartComponents/SystemDetail/InventoryDetail.js
adonispuente: Why is that? The route has` :inventoryId `inside of it and this just grabs that arg from the url. 
In the scenarios I can think of where the url can be messed up, it would break inventory in the old version as well

src/SmartComponents/AdvisoryDetail/__snapshots__/AdvisoryDetail.test.js.snap
adonispuente: This still needs to be a 'to', but without changing the paths.path to a hardcoded string, I wasnt able to find a way where this wasnt angry about this field. It seems like it never 'gets passed down' everything/ it cant get the Route info the same as before. 

To try to elaborate, with the new set up, routes sort of render dynamically, which is why we're able to do something like paths.path in the actual js files and it understands which path to look at. I dont think the test work the same way(I could be wrong here but this seems to be to case). The test just grab the entire path array and looks for paths.path which of course fails.

src/SmartComponents/PatchSet/PatchSetAssets.js
```diff
@@ -1,14 +1,23 @@
...
+    EllipsisVIcon
+} from '@patternfly/react-icons';
 import { intl } from '../../Utilities/IntlProvider';
 import messages from '../../Messages';
 
-export const createPatchSetButton = (setPatchSetState) => () =>
-    (<Button key='createButton' onClick={() => setPatchSetState({ isPatchSetWizardOpen: true })}>
-        {intl.formatMessage(messages.labelsButtonCreateTemplate)}
-    </Button>);
+export const createPatchSetButton = (setPatchSetState, hasAccess) => () =>
```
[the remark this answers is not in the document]
adonispuente: If this is capitalized, in PatchSet,  I have to import it as 
`CreatePatchSetButton as createPatchSetButton`
or it'll yell about me needing to use 'functions with a capital letter can only be called as a constructor'.

src/PresentationalComponents/WithPermission/WithPermission.js
```diff
@@ -0,0 +1,20 @@
+import React  from 'react';
```
mkholjuraev, line 1: Nice to have later: This component together with PermissionRouter can be pulled into FEC as shared and then used across applications.

src/SmartComponents/AdvisoryDetail/AdvisoryDetail.js
```diff
@@ -51,7 +51,7 @@ const AdvisoryDetail = ({ match }) => {
...
+                        to: paths.path,
```
mkholjuraev, line 54: `paths` is an array in Router.js. This looks like a mistake that you are treating it as an object here.

src/SmartComponents/PackageDetail/PackageDetail.js
```diff
@@ -49,7 +49,7 @@ const PackageDetail = ({ match }) => {
                 breadcrumbs={[
                     {
                         title: intl.formatMessage(messages.titlesPatchPackages),
-                        to: paths.packages.to,
+                        to: paths.path,
```
mkholjuraev, line 52: Here also, it should be an array?

adonispuente:
@bastilian @mkholjuraev Regarding the breadcrumbs, it seems passing in paths.path is broken. 
It will resolve by either having a hard string, are by slapping in something like paths[0].path
You've both had different suggestions for both so I'd like to come to an agreement.

I actually prefer having strings in there, or I can make constants and put them there as well if thats something you guys would prefer

mkholjuraev:
@adonispuente I also opt for the string rather than paths[0].path.

mkholjuraev (approved):
LGTM! tested locally using account `psegedy_patch` which has only read permission. For accounts with full permission, everything seems to be as described.

PR #910, opened 2022-11-09
Select only ids

mkholjuraev:
Resolves: [SPM-1783](https://issues.redhat.com/browse/SPM-1783).

Generally, this PR introduces using new endpoints to select only ids instead of full data. This has optimized the time to fetch those ids. Apart from that, the PR has some refactoring job that is essentially moving useOnSelect hook into a separate file, improving readability and structure.


adonispuente (changes requested):
Currently if you add filters to the table and begin to select all, it ignores the filters and still selects all available systems. The useOnSelect is ignoring filters
Ex. If I have 1000 total systems and filter down to 200, when I 'select all', it will select all 1000

mkholjuraev:
@adonispuente thank you for the review. Good catch, fixed.

adonispuente (approved):
PR works as intended, LGTM!

PR #917, opened 2022-11-18
Create pull_request_template.md

mkholjuraev:
Creates a PR template. Feel free to suggest improvements.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -7,7 +7,7 @@ Please include a summary of the change, what this fixes/creates/improves.
 
 # How to test the PR
 
-Please inlcude steps to test your PR.
+Please include steps to test your PR.
 
 # Before the change
 
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,14 @@
+# Description
+
+Please include a summary of the change and which issue is fixed. .
+
+Associated Jira ticket: # (issue)
+
+# How to test the PR
+
+
+# Checklist:
+
+- [ ] The commit message has the Jira ticket linked
+- [ ] Tests for the changes have been added
+- [ ] Read.me is updated if necessary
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,24 @@
...
+# After the change
+
+
+# Checklist:
+
+- [ ] The commit message has the Jira ticket linked
+- [ ] PR has a short description
+- [ ] Screenshots before and after the change are added
+- [ ] Tests for the changes have been added
+- [ ] Read.me is updated if necessary
```
bastilian, line 24: Maybe as an optional point in the checklist/testing requirements, we could add eventual dependencies. For example when something needs a PR from frontend-components or it needs to be run with another app together.

.github/pull_request_template.md
```diff
@@ -0,0 +1,24 @@
+# Description
+
+Associated Jira ticket: # (issue)
```
gkarat, line 3: Can we agree on having a full link? Going to a board and copy-pasting the id is slower that one click 😅

.github/pull_request_template.md
gkarat, line 24:
```suggestion
- [ ] README.md is updated if necessary
```

gkarat (approved):
LGTM

.github/pull_request_template.md
bastilian: Though do we really need it in the PR? If the issue ID has the right format and is in the commit message, GitHub should turn it into a link on the commit view. That is at least how I always go to Jira issues from PRs.

.github/pull_request_template.md
mkholjuraev: @bastilian it makes sense. What would be the best wording for this?

.github/pull_request_template.md
gkarat: yes, in PR it is not strictly needed (I also look at commits, not PRs), let's mention only URLs in commits

PR #918, opened 2022-11-23
chore(RHCLOUD-23128): Remove get registry dependency and use create store directly

karelhala:
# Description

Associated Jira ticket: #RHCLOUD-23128

The function `getRegistry` is no longer needed, on contrary with module federation it can bring additonal problems because of incorrect access of a store instance. This PR removes the usage of said function and replaces it with proper `store.replaceReducer` function.

# How to test the PR

Run patch locally and test if everything works as before (pay closer attention to pages with inventory in them).

### DIFF
--- a/src/store/index.js
+++ b/src/store/index.js
@@ -1,6 +1,5 @@
 import notificationsMiddleware from '@redhat-cloud-services/frontend-components-notifications/notificationsMiddleware';
 import { notificationsReducer } from '@redhat-cloud-services/frontend-components-notifications/redux';
-import { getRegistry } from '@redhat-cloud-services/frontend-components-utilities/Registry';
 import promiseMiddleware from 'redux-promise-middleware';
 import { AdvisoryDetailStore } from './Reducers/AdvisoryDetailStore';
 import { AdvisoryListStore } from './Reducers/AdvisoryListStore';
@@ -16,9 +15,9 @@ import { AdvisorySystemsStore } from './Reducers/AdvisorySystemsStore';
 import { GlobalFilterStore } from './Reducers/GlobalFilterStore';
 import { PatchSetsReducer } from './Reducers/PatchSetsReducer';
 import { SpecificPatchSetReducer } from './Reducers/SpecificPatchSetReducer';
+import { legacy_createStore as createStore, applyMiddleware, combineReducers } from 'redux';
 
-const registry = getRegistry({}, [promiseMiddleware, notificationsMiddleware()]);
-registry.register({
+export const defaultReducers = {
     AdvisoryListStore,
     SystemDetailStore,
     SystemAdvisoryListStore,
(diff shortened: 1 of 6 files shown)


src/store/index.js
```diff
@@ -16,9 +15,9 @@ import { AdvisorySystemsStore } from './Reducers/AdvisorySystemsStore';
 import { GlobalFilterStore } from './Reducers/GlobalFilterStore';
 import { PatchSetsReducer } from './Reducers/PatchSetsReducer';
 import { SpecificPatchSetReducer } from './Reducers/SpecificPatchSetReducer';
+import { legacy_createStore as createStore, applyMiddleware, combineReducers } from 'redux';
```
gkarat, line 18: I saw 4.2.0 has deprecated createStore. Can we then explicitly set https://github.com/RedHatInsights/patchman-ui/blob/master/package.json#L29 to "4.1.8" - the version before? So that we are not using legacy_* like functions. We can also consider utilizing redux-toolkit as suggested by redux devs: https://github.com/reduxjs/redux/releases/tag/v4.2.0

mkholjuraev: We would need to migrate to the redux toolkit progressively and I am afraid downgrading redux version might lead to other problems.

karelhala: I don't like redux-toolkit. They added so much complexity when creating redux store it's unbelievable. `react-redux` team won't be removing the createStore function.

```
You should not be using the `redux` core package by itself today, except for learning purposes.
 The `createStore` method from the core `redux` package will not be removed, but we encourage
 all users to migrate to using Redux Toolkit for all Redux code.
 
If you want to use `createStore` without this visual deprecation warning, use
the `legacy_createStore` import instead:

`import { legacy_createStore as createStore} from 'redux'`
```

gkarat: @mkholjuraev, I don't think we are downgrading in fact :-) https://github.com/RedHatInsights/patchman-ui/blob/master/package.json#L29: here the version is actually set to even older one

gkarat: @karelhala, wow, then I am really curious what exactly made you problems. I remember setting up the store using this library for both Advisors: I had an absolutely opposite experience... Still, I would like to see the bundle difference when using `redux` and `redux-toolkit` - probably the only thing I am concerned about in regards to this library.

mkholjuraev: > @mkholjuraev, I don't think we are downgrading in fact :-) https://github.com/RedHatInsights/patchman-ui/blob/master/package.json#L29: here the version is actually set to even older one

Aha, I meant the version in the PR. As you know each bug fix release has something to be careful of that is fixed in the next release.

karelhala: @gkarat the bundle size is fine. It's just the configuration itself. In redux-toolkit they are using a lot of magical functions to build something as simple as redux store, it's just a state machine and the redux team is putting a lot of fancy functions to make it "easier" for folks to understand. One example is immutability, if you understand redux properly you know you can't change the state from outside so you don't need to have any immutability library to work with it. But that's my personal opinion and as @mkholjuraev mentioned, we'd have to rewrite a bunch of code (in patch) to use it.

mkholjuraev (approved):
LGTM! Works as expected. I do not see any issues while testing. Just to be sure, I checked against recent reload issues, and I did not see it happening. Thanks

PR #919, opened 2022-11-23
fix(SPM-1768): Adds custom role to templates page

adonispuente:
# Description

Associated Jira ticket: # (SPM-[1768](https://issues.redhat.com/browse/SPM-1768))

This adds a new role to the allow list of roles that can create/edit patch templates. Admin has access to everything, and I had set it so that only admin can perform template actions. The custom role should have access too it as well. This change allows both admin and a user with patch:template:write permissions to perform template actions

# How to test the PR

Navigate the Templates page with an account that has patch:template:write access and read permissions, and verify that you can create and edit a template
Do the same for an account with admin permissions
Verify that if a user with read permissions goes to the page they dont have access to the button and cant edit the templates.

If you need access to accounts with these permission's, please let me know
# Before the change

The custom role didnt work

# After the change
The custom role works

# Checklist:

### DIFF
--- a/src/SmartComponents/PatchSet/PatchSet.js
+++ b/src/SmartComponents/PatchSet/PatchSet.js
@@ -117,7 +117,8 @@ const PatchSet = ({ history }) => {
     };
 
     const { hasAccess } = usePermissionsWithContext([
-        'patch:*:*'
+        'patch:*:*',
+        'patch:template:write'
     ]);
     const CreatePatchSetButton = createPatchSetButton(setPatchSetState, hasAccess);
     const actionsConfig = patchSetRowActions(openPatchSetEditModal, handlePatchSetDelete);


mkholjuraev:
@adonispuente users with admin rights are not checked by RBAC, rather it is checked by the platform/engine. Am I right? I want to understand how users with admin rights are checked.

adonispuente:
@mkholjuraev The answer is both RBAC and the engine check for it. Its like having 2 layers of security. Its essential to have it on the backend (at-least thats what I think if you're only going to pick one), but on the front-end (taking the templates page for example) if a user with invalid access was somehow able to send a POST request and send a template, it would be denied by the engine. Having it checked on the frontend just allows you to do things like this where a user can see the page without having the ability to make any templates. It really just helps add more options to give a user to experience the app.

mkholjuraev:
And now the PR looks good with `'patch:*:*'`. With this, we are checking admin users. Approving

mkholjuraev (approved):
LGTM! @adonispuente  thanks

PR #930, opened 2022-12-08
Optimizations

mkholjuraev:
# Description

for optimization

# How to test the PR

Please play around the UI.

# Before the change

# After the change

# Dependent work link

# Checklist:


src/App.js
```diff
@@ -39,7 +29,7 @@ const App = () => {
 
                 const globalFilterConfig = mapGlobalFilters(TAGs, SIDs, data?.Workloads);
 
-                if (!isEqual(config, globalFilterConfig)) {
+                if (JSON.stringify(config) === JSON.stringify(globalFilterConfig)) {
```
gkarat, line 32: I am again using `JSON.stringify` for deep comparison. For example, `JSON.stringify({a: 42, b: 0}) === JSON.stringify({b: 0, a: 42})` results to false. https://stackoverflow.com/questions/201183/how-to-determine-equality-for-two-javascript-objects#comment82819893_46645497

Lodash is making this job perfectly.

src/App.js
```diff
@@ -1,31 +1,21 @@
-import NotificationPortal from '@redhat-cloud-services/frontend-components-notifications/NotificationPortal';
-import '@redhat-cloud-services/frontend-components-notifications/index.css';
-import { isEqual } from 'lodash';
```
gkarat, line 3: If you ever use lodash functions again, please consider the import from the function directory like
```
import isEqual from 'lodash/isEqual';
```
This will let you avoid importing the whole lodash library into your js bundles.

gkarat (changes requested):

src/SmartComponents/PackageSystems/PackageSystems.js
```diff
@@ -83,10 +83,10 @@ const PackageSystems = ({ packageName }) => {
         ]
     };
 
-    const activeFiltersConfig = {
+    const activeFiltersConfig = useMemo(() => ({
         filters: buildFilterChips(filter, search, intl.formatMessage(messages.labelsFiltersSystemsSearchTitle)),
         onDelete: deleteFilters
-    };
+    }), []);
```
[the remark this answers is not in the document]
mkholjuraev: Ups, I forgot to add the deps. Will add them

src/App.js
mkholjuraev: The expected false in your example is what I want. The idea behind deep comparisons is also somehow avoiding this extra trigger. I am interested in the value pairs not their order in the object.

src/App.js
mkholjuraev: You are giving comments about what I got rid of already :)

gkarat: yep, I was rather referring to my previous comment where I suggested to use lodash for deep comparisons. and it implied this comment :-P

src/App.js
gkarat: aha, in that case this should be fine. (though you are technically adding another requirement for this variable to be always the same order; this can complicate refactor/debugging in the future)

gkarat (approved):
this looks good! if it's considered complete, I am OK with merging it.

just a side nit-pick: I would like to make some profiling before/after such PRs to see whether how much this increases the performance. some of the `useCallback`s in tiny, not huge components make me wonder if it really provides any value to add it there. e.g., here is well written article about the different use cases for useCallback: https://dmitripavlutin.com/react-usecallback.

mkholjuraev:
@gkarat thank you for the review! I have skimmed through the doc you have linked. Indeed, it has good examples.

While working on the optimizations, I also referred to profiling and measured the re-renders. It seems that most of the re-renders are necessary re-renders together with platform dependant ones. Thus, there is not much saving on re-renders with this PR unfortunately :). But still, there are on average ~15 re-renders on the first load, thus trying to avoid such a tiny function recreation can be a good thing in the end. 

Regarding the **useCallback** in the searchFilter, this is to get rid of keeping a debounced function in a state. This was wrong, the correct way is to store it in a useCallback to have only one instance across re-renders.

PR #978, opened 2023-03-02
feat: Table links and sorting fix

leSamo:
# Description
- Link from Advisory detail to Template detail page
- Fix sorting on Template detail page
- Link from Template detail to Systems detail page

## Testing instructions
a) Go to Advisory detail page and in the table find a system that has a template. Verify you can navigate to Template detail page by clicking on the system name inside the table.
b) In Template detail page you should be able to correctly sort the table by display name. All other columns should have sorting indicators disabled.
c) Go to Template detail page and click on any system name inside the first column. You should be correctly navigated to Systems detail page of clicked system.


bastilian (approved):
LGTM! Thank you @leSamo!

PR #979, opened 2023-03-03
Template system unassignment from Template detail page

leSamo:
# Description

Associated Jira ticket: [SPM-1904](https://issues.redhat.com/browse/SPM-1904), [SPM-1907](https://issues.redhat.com/browse/SPM-1907), [SPM-1927](https://issues.redhat.com/browse/SPM-1927)

From System list page:
It was not possible to remove template from stale systems. This is now fixed.

From Template detail page:
Add table row actions accessible with patch:template:write permission, these actions contain action to unassign system from a template.

## Testing instructions

**a) System list page:**
Find a system that has template assigned.

i) Unassign system from template using row actions kebab.
ii) Unassign multiple systems from template using bulk action kebab in the table toolbar. Make sure to try to select some systems that do have template and some that don't - modal should then point out to you that some systems don't have a template and won't be affected.

**b) Template detail page:**
Unassign system from template using row actions kebab.
Bulk action is not currently implemented.

### DIFF
--- a/src/SmartComponents/Modals/Helpers.js
+++ b/src/SmartComponents/Modals/Helpers.js
@@ -5,7 +5,10 @@ import messages from '../../Messages';
 import { fetchSystems } from '../../Utilities/api';
 
 export const filterSystemsWithoutSets = (systemsIDs) =>  {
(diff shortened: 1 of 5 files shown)


bastilian (approved):
Works as expected! Thank you @leSamo!

PR #981, opened 2023-03-07
Remove template docs links

leSamo:
As per PM's request (https://redhat-internal.slack.com/archives/CPS5FH07Q/p1678202549907769?thread_ts=1678119878.115079&cid=CPS5FH07Q) we are, for now, removing all documentation link related to templates.

These places include:
- Template list page empty state (displayed when user has 0 templates)
- Template list page title tooltip
- Template creation/editing wizard just below the wizard title


mkholjuraev:
Maybe we can use a feature flag?

PR #988, opened 2023-03-10
Fix Template detail table advisory fallback text

leSamo:
The messages for corresponding columns were accidentally swapped.

## Before:
Screenshot from 2023-03-10 12-07-15

## After:
Screenshot from 2023-03-10 12-06-52


bastilian (approved):
Thank you for fixing this @leSamo!

PR #989, opened 2023-03-10
Add 'Creator' filter to Template list page

leSamo:
# Description

Associated Jira ticket: [SPM-1932](https://issues.redhat.com/browse/SPM-1932)

Add 'Creator' checkbox filter to Template list page. The options are sent from the backend inside meta section of API response and then saved inside Template list store.

Screenshot from 2023-03-10 13-16-38


bastilian (approved):
Thank you @leSamo! Works as expected!

PR #990, opened 2023-03-10
Decouple step title and sidebar name in Template Wizard

leSamo:
# Description

Associated Jira ticket: [SPM-1893](https://issues.redhat.com/browse/SPM-1893)

- until now the title of the step had to be the same for header label and sidebar label, but in the future we will need them to be different so I decoupled these strings by removing `showTitles: true` and manually adding header to each step
- update System selection step text to the [latest mockups](https://www.sketch.com/s/598ac9a1-16ce-47b1-8515-e79928a2b2fb/a/eKRw04p)
- if user leaves description field empty, review step will state 'No description provided' instead of 'No description available'

### DIFF
--- a/src/Messages.js
+++ b/src/Messages.js
@@ -700,7 +700,7 @@ export default defineMessages({
     textTemplateReviewSystems: {
         id: 'textTemplateReviewSystems',
         description: 'text for patch template wizard',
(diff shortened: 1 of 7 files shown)


bastilian (approved):
LGTM! Thank you @leSamo!

PR #991, opened 2023-03-14
Template detail empty state button launches wizard

leSamo:
# Description

Associated Jira ticket: [SPM-1893](https://issues.redhat.com/browse/SPM-1893)

- if template has 0 systems it shows an empty state on template detail page, the button in the empty state now correctly opens wizard; previously it didn't do anything
- removed unused 'assign' template wizard type
- editing a template from template detail page now has correct "edit" verb instead of "create" in the titles


bastilian (approved):
Wizard now opens as expected! Thank you @leSamo!

PR #995, opened 2023-03-17
Fix numerous prop types warning

leSamo:
Fix prop types warnings in the developer console.
There should be no change in functionality.


gkarat (approved):
LGTM, can't see any issues/visual differences while running locally

PR #997, opened 2023-03-21
[RHIF-128] adopt useChrome hook

johnsonm325:
# Description

# How to test the PR

# Before the change

# After the change

# Dependent work link

# Checklist:


src/Utilities/axiosInterceptors.js
```diff
@@ -7,16 +7,6 @@ import { ReadOnlyNotification } from './constants';
 const axiosInstance = axios.create();
 
 export async function  accessCheckInterceptor (config)  {
```
mkholjuraev, line 9: You can remove the whole interceptor

src/Routes.js
```diff
@@ -89,77 +90,80 @@ const TemplateDetail = lazy(() =>
...
-];
-
 export const Routes = () => {
     const history = useHistory();
+    const chrome = useChrome();
+
+    const isPatchSetEnabled = useFeatureFlag(featureFlags.patch_set);
+    const generalPermissions = ['patch:*:*', 'patch:*:read'];
+
+    const paths = useMemo(() => {
```
mkholjuraev, line 100: I think this hook can be extracted into a separate file for cleanness

mkholjuraev: As, this work needs to be released ASAP let's ignore this.

mkholjuraev (approved):

src/Utilities/axiosInterceptors.js
```diff
@@ -6,20 +6,6 @@ import { ReadOnlyNotification } from './constants';
 
 const axiosInstance = axios.create();
 
-export async function  accessCheckInterceptor (config)  {
```
bastilian, line 9: Removing this interceptor causes some fetches now to be called, and the promises that these fetches are returning are not awaited for. This is the reason for the failing tests. 

The resolution to this is likely going to be to inspect all store actions that do async fetches and make sure they are properly awaited for. Maybe there is an easier way, but we might need to dig a bit more for that.

mkholjuraev (changes requested):
When I run this PR, I am getting broken page. 

image

Fewwy:
I tested the PR and it works for me. LGTM

PR #1000, opened 2023-03-24
Fix template up-to date being off by a day in negative UTC timezones

leSamo:
- Fix template up-to date being off by a day in negative UTC timezones
- add hasGutter to header so there's margin between long template names and dropdown, after screenshot:
Screenshot from 2023-03-24 06-43-08

How to test:
1. Set your timezone to some place with negative UTC offset, for example New York. 
2. Create a template with some date.
3. In the last (review) step of the template creation check if the date is the same as you set.
4. Go to detail page of your just created template.
5. Check if the date in the header is the same as you set.
6. Don't forget to set your timezone back :grinning:

### DIFF
--- a/src/Utilities/Helpers.js
+++ b/src/Utilities/Helpers.js
@@ -602,6 +602,25 @@ export const convertIsoToDate = (isoDate) => {
         `-${dateObject.getDate().toString().padStart(2, '0')}`;
 };
 
+// 2023-03-05 -> 05 Mar 2023
+// 2023-03-22T20:00:00-04:00 (ISO format) -> 23 Mar 2023
+export const templateDateFormat = (dateString) => {
+    if (!dateString) {
+        return 'N/A';
+    }
+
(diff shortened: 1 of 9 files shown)


src/Utilities/Helpers.js
```diff
@@ -602,6 +602,25 @@ export const convertIsoToDate = (isoDate) => {
         `-${dateObject.getDate().toString().padStart(2, '0')}`;
 };
 
+// 2023-03-05 -> 05 Mar 2023
+// 2023-03-22T20:00:00-04:00 (ISO format) -> 23 Mar 2023
+export const templateDateFormat = (dateString) => {
```
mkholjuraev, line 607: One unit test would make this even more robust 👍

mkholjuraev (approved):
LGTM! @leSamo thank you.

src/Utilities/Helpers.js
leSamo: Very good suggestion! I just added some :grinning:

PR #1014, opened 2023-04-03
Fix date-to resetting when assigning a system to template

leSamo:
There was a bug when you assigned a system to a template in the system list table, the template would have the date-to reset to January 1, year 1. This has now been fixed.

### DIFF
--- a/src/Utilities/Hooks.js
+++ b/src/Utilities/Hooks.js
@@ -261,7 +261,7 @@ export const usePatchSetApi = (wizardState, setWizardState, patchSetID) => {
 
     const onSubmit = React.useCallback((formValues) => {
         const { name, description, toDate, id } = formValues.existing_patch_set || formValues;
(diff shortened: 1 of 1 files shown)


src/Utilities/Hooks.js
```diff
@@ -269,7 +269,7 @@ export const usePatchSetApi = (wizardState, setWizardState, patchSetID) => {
             name,
             description,
             inventory_ids: (patchSetID || id) ? objUndefinedToFalse(systems) : Object.keys(systems),
-            config: { to_time: fomattedDate }
+            ...fomattedDate ? { config: { to_time: fomattedDate } } : {}
```
adonispuente, line 272: This works, but I would recommend something like 
`...fomattedDate && { config: { to_time: fomattedDate } }`

adonispuente (approved):
Small nitpick but not completely necessary. LGTM!

PR #1016, opened 2023-04-03
Fix Advisory list page applicable system column

leSamo:
See comment in [SPM-1948](https://issues.redhat.com/browse/SPM-1948)

- rename "Applicable systems" to "Affected systems" in table and dashbar
- update `advisories/` endpoint version from v2 to v3 to fix incorrect count of systems in applicable/affected systems column

### DIFF
--- a/src/Messages.js
+++ b/src/Messages.js
@@ -19,7 +19,7 @@ export default defineMessages({
     labelsApplicableSystemsCount: {
         id: 'labelsApplicableSystemsCount',
         description: 'applicable systems number label',
-        defaultMessage: '{systemsCount} applicable systems'
+        defaultMessage: '{systemsCount} affected systems'
     },
     labelsBulkSelectAll: {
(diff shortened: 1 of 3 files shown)


src/Messages.js
```diff
@@ -19,7 +19,7 @@ export default defineMessages({
     labelsApplicableSystemsCount: {
         id: 'labelsApplicableSystemsCount',
         description: 'applicable systems number label',
-        defaultMessage: '{systemsCount} applicable systems'
+        defaultMessage: '{systemsCount} affected systems'
```
adonispuente, line 22: Just a nitpick, but this would be better as a new message. The title of it is  labelsApplicableSystemsCount but the message says affected systems. By this I mean keep this message here for keepsake and make a new message called labelsAffectedSystemsCount

adonispuente (approved):
LGTM!
Just that small nitpick.

PR #1020, opened 2023-04-11
Update all remaining API requests to use v3

leSamo:
There should be no change in functionality.


mkholjuraev (approved):
LGTM codewise! not testing locally as the changes are obvious.

PR #1021, opened 2023-04-11
Add status column to Advisory systems table

leSamo:
Part of [SPM-1949](https://issues.redhat.com/browse/SPM-1949)

1. Add status column to Advisory systems table
2. If status column value is 'Applicable' the row checkbox is disabled


Fewwy:
Tested it locally, LGTM

Fewwy (approved):

PR #1022, opened 2023-04-12
Fix adding/removing template not causing refresh in System detail header

leSamo:
Associated Jira ticket: [SPM-2003](https://issues.redhat.com/browse/SPM-2003)


mkholjuraev (approved):
LGTM!

PR #1027, opened 2023-04-17
Add status column to System detail Advisory table

leSamo:
https://issues.redhat.com/browse/SPM-1944

To test:
- Status column was added to System detail Advisory table
- Status column is sortable
- Iff advisory status is "Applicable", the selection checkbox should be disabled


Fewwy (approved):
LGTM

johnsonm325 (approved):

PR #1029, opened 2023-04-19
Fix template modal system table not resetting to page 1

leSamo:
See https://github.com/RedHatInsights/patchman-ui/pull/1028#pullrequestreview-1391792579

Bug that was fixed happened in Template modal -> System review step if you navigated to page 2 and then applied filter which returned less than 20 rows; calling page 2 on this data caused 400 error. Fixed by resetting to page 1 on every filter change, which is behaviour of all other Patchman tables (and Vulnerabilities as well).


AsToNlele (approved):
LGTM good find :smile:

PR #1031, opened 2023-04-21
Update System review step of Template wizard

leSamo:
Mockups: https://www.sketch.com/s/598ac9a1-16ce-47b1-8515-e79928a2b2fb/a/DP7Ka9y

- Shorten the first sentence according to mockups
- Move the second sentence inside a warning alert, not exactly according to mockups, but full implementation is blocked, so this is a workaround
- Add "Last updated" column to the table according to mockups

## Before:
Screenshot from 2023-04-22 00-39-19

## After:
Screenshot from 2023-04-22 00-38-47


AsToNlele (approved):
LGTM! :smile:

PR #1040, opened 2023-05-02
(RHIN-830): Add Zero State to Patch

adonispuente:
As in accordance to these mocks, this PR adds zero state to patch,  https://docs.google.com/document/d/1FGuzbL2jyRmn_coYD5Z_o3d0MhOf1vQDT46NQvsfPyA/edit#heading=h.1hvxoje16w3i  https://www.sketch.com/s/550bd962-ca1f-42d8-8d89-d457527a059a/a/qbe9Ol2#Inspect   https://www.sketch.com/s/550bd962-ca1f-42d8-8d89-d457527a059a/prototype/a/AEF47CA9-8F00-40FC-B39B-3B989D3F83FD   To test, sign into an account with 0 systems connected

### DIFF
--- a/src/Routes.js
+++ b/src/Routes.js
@@ -1,5 +1,5 @@
 import PropTypes from 'prop-types';
-import React, { lazy, Suspense, useCallback, useEffect } from 'react';
+import React, { lazy, Suspense, useCallback, useEffect, useState } from 'react';
 import { Redirect, Route, Switch, useHistory } from 'react-router-dom';
 import WithPermission from './PresentationalComponents/WithPermission/WithPermission';
 import { Bullseye, Spinner } from '@patternfly/react-core';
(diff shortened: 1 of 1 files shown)


mkholjuraev (changes requested):
Patch has a good handler implemented for the zero (empty) state [here](https://github.com/RedHatInsights/patchman-ui/blob/master/src/PresentationalComponents/Snippets/ErrorHandler.js#L53). Patch api adds has_systems boolean to check this senario. 
let's make use of it by importing the zero-state module inside [NoRegisteredSystems](https://github.com/RedHatInsights/patchman-ui/blob/master/src/PresentationalComponents/Snippets/NoRegisteredSystems.js) component

mkholjuraev (approved):
As we have a short time, let's merge this and work on refactoring this not to use extra API calls to /hosts later after the code freeze.

Tested locally, LGTM. The only thins that caught my attention is that the global filter is hidden in the mocks but not in this PR.

PR #1042, opened 2023-05-02
Check if template name exists when creating a template

leSamo:
https://issues.redhat.com/browse/SPM-1952

Inside the template wizard:
- if user enters template name which is already used, they cannot continue to the next step
- if user enters template name which does not match regex `^[a-z0-9][a-z0-9_-]*$`, they cannot continue to the next step


gkarat (approved):
LGTM! Tried locally under stage-preview environment. Haven't spotted any code issues

PR #1044, opened 2023-05-05
Add template assign modal

leSamo:
Add a new modal when assigning systems to template from System list page

Screenshot from 2023-05-05 03-11-24

Mockups: https://www.sketch.com/s/598ac9a1-16ce-47b1-8515-e79928a2b2fb/a/Om3zKYq#Inspect

So far it's not 100% according to the mockups, but it's good a large step in the right direction.

PR #1045, opened 2023-05-05
Add template assignment modal to system detail page

leSamo:
Same as #1044 but for System detail page when you click `Assign to a template` from header dropdown


AsToNlele (approved):
LGTM :smile:

PR #1049, opened 2023-05-09
[WIP] Adding Patch smoke ui test config pr_check.sh

niyazRedhat:


pr_check.sh
```diff
@@ -12,26 +12,38 @@ export APP_ROOT=$(pwd)
...
-# IQE_MARKER_EXPRESSION=""
-# IQE_FILTER_EXPRESSION=""
+export DEPLOY_FRONTENDS="true"
+export IQE_ENV="ephemeral"
+export IQE_SELENIUM="true"
+export APP_NAME="patchman"
+export DEPLOY_TIMEOUT="900"  # 15min
+export IQE_CJI_TIMEOUT="30m"
+export REF_ENV="insights-stage"
+export COMPONENTS_W_RESOURCES="vmaas"
```
mtclinton, line 34: I think you need to combine these to be: `export COMPONENTS_W_RESOURCES="vmaas patchman"` otherwise the vmaas one is overwritten. See here:  https://github.com/RedHatInsights/vulnerability-engine/blob/master/pr_check.sh#L9C19-L9C19

niyazRedhat: @mtclinton thanks for looking into it. i tried your above suggestion. Vmaas works fine.
raised another PR https://github.com/RedHatInsights/patchman-ui/pull/1077
now vmaas looks good to me :)

niyazRedhat:
raised another PR for the same https://github.com/RedHatInsights/patchman-ui/pull/1077

PR #1055, opened 2023-05-17
chore(deps): upgrade react-router-dom to v6

AsToNlele:
# Description

Associated Jira ticket: [SPM-1858](https://issues.redhat.com/browse/SPM-1858)

How to test:
1. Make sure all the links redirect correctly
2. Approve
3. Profit?

# Checklist:

- [x] Tests for the changes have been added


AsToNlele:
I'll edit the description steps later

AsToNlele:
After Frontend components have been upgraded, there is one less step :tada:

src/Routes.js
```diff
@@ -196,36 +116,39 @@ export const Routes = () => {
...
-                        (
-                            (!some(paths, p => p.to === history.location.pathname)) && (
-                                <Redirect to={'/advisories'} />
-                            )
-                        )
-                    }
-                    />
-                </Switch>
-            }
+            <Router>
```
bastilian, line 119: Something is not right here. This should throw an error like "Router in Router not allowed".

bastilian: 🙈 my bad. I didn't see the `as` for the import. But maybe we can avoid that renaming. What if we use `AdvisorRoutes` for this const and not rename the `Routes` component from react router?

bastilian (approved):
Nice job! Thank you @AsToNlele!

bastilian:
@AsToNlele Just out of precaution, we should get a second approval from @RedHatInsights/team-interact before merging.

bastilian (changes requested):
Sorry, but I did find issues doublechecking on `/preview`.
There everything appears to fall apart. :/

src/Routes.js
```diff
@@ -91,81 +74,18 @@ const TemplateDetail = lazy(() =>
...
-            component: Templates
-        }
-    ];
-
     const listenNavigation = useCallback(() => {
         if (chrome) {
             return chrome.on('APP_NAVIGATION', event => {
                 if (event.domEvent) {
-                    history.push(`/${event.navId}`);
+                    navigate(`/${event.navId}`);
```
bastilian, line 88: I think, this should use `chrome.appNavClick` like here[0] in order to work correctly. We have encountered with navigate where it does a full page refresh, which it shouldn't.

[0] https://github.com/RedHatInsights/compliance-frontend/pull/1892/files#diff-3d74dddefb6e35fbffe3c76ec0712d5c416352d9449e2fcc8210a9dee57dff67R23

johnsonm325 (approved):
This seems to work okay.

There is a bug when you apply an os filter to the systems table and then try to refresh the page. But it exists in stage, so isn't related to this PR. I'm creating a card.

src/Routes.js
```diff
@@ -90,82 +73,18 @@ const TemplateDetail = lazy(() =>
...
-            component: Templates
-        }
-    ];
-
     const listenNavigation = useCallback(() => {
         if (chrome) {
             return chrome.on('APP_NAVIGATION', event => {
                 if (event.domEvent) {
-                    history.push(`/${event.navId}`);
+                    chrome.appNavClick(`/${event.navId}`);
```
bastilian, line 87: According to the platform team, this whole part related to `APP_NAVIGATION` can and should be removed (see https://github.com/RedHatInsights/compliance-frontend/pull/1892/files#diff-3d74dddefb6e35fbffe3c76ec0712d5c416352d9449e2fcc8210a9dee57dff67 for example)

PR #1057, opened 2023-05-19
Bulk unassign systems from template detail page

leSamo:
https://issues.redhat.com/browse/SPM-1943


gkarat (approved):
tested locally. the bulk select is now visible, and I could remove more systems. lgtm, thank you 👍🏼

PR #1058, opened 2023-05-19
Add fallback to OS column on Template detail page

leSamo:
If template system had OS data unavailable it only showed the blue icon. Correct behaviour is to show "N/A" and no blue icon. 

### Before:
Screenshot from 2023-05-19 16-02-01

### After:
Screenshot from 2023-05-19 16-02-03

### How to test:
- go to https://stage.foo.redhat.com:1337/beta/insights/patch/templates/8264?page=4&perPage=20&sort=os
- observe OS column contents


gkarat (approved):

gkarat:
tested locally, the issue is resolved, thank you

PR #1062, opened 2023-05-25
Fix stale system count in Dashbar

leSamo:
https://issues.redhat.com/browse/SPM-2054

### After
Screenshot from 2023-05-25 12-16-06

- stale system count used to display 0 instead of the correct amount, this was fixed by adding staleness param to the API request
- improved the responsivity of the dashbar
- fixed camel case `SystemsStatusreport` -> `SystemsStatusReport`


Fewwy (approved):
Fixed! LGTM

PR #1067, opened 2023-05-30
Include only installable advisories when remediating all

leSamo:
https://issues.redhat.com/browse/SPM-2069

1. Go to system list page
2. Find a system with a template, that has some applicable (non-installable) advisories
3. In row actions menu select "Apply all applicable advisories"
4. Observe how many issues are selected -- previously it selected applicable and installable advisories, now it should only select installable

### DIFF
--- a/src/SmartComponents/Systems/SystemListAssets.test.js
+++ b/src/SmartComponents/Systems/SystemListAssets.test.js
@@ -34,7 +34,8 @@ describe('SystemListAssets.js', () => {
         it('Should call fetchApplicableSystemAdvisoriesApi with row id and limit: 10000', () => {
             expect(fetchApplicableSystemAdvisoriesApi).toHaveBeenCalledWith({
                 id: 'testId',
-                limit: 10000
+                limit: -1,
+                'filter[status]': 'in:Installable'
             });
         });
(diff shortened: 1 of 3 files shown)


mkholjuraev (approved):
LGTM! Happy to see this

src/SmartComponents/Systems/SystemListAssets.test.js
```diff
@@ -34,7 +34,8 @@ describe('SystemListAssets.js', () => {
         it('Should call fetchApplicableSystemAdvisoriesApi with row id and limit: 10000', () => {
             expect(fetchApplicableSystemAdvisoriesApi).toHaveBeenCalledWith({
                 id: 'testId',
-                limit: 10000
+                limit: -1,
```
psegedy, line 37: @leSamo @mkholjuraev guys use pagination if possible, this way if there are 30k systems it will for sure time out on backend, the reasonable page size is 100 (maybe 1000 with fast apis)

psegedy: /views/systems/advisories with system id in POST body should be the best way how to do it IMHO, you don't need to deal with pagination and you will get all advisory ids for a single system

PR #1070, opened 2023-05-31
Ls assets

mkholjuraev:
This enables pushing assets to sentry. This also removes sources folder from ./dist to prevent too much bundle.


adonispuente (approved):
Lets test it out!

PR #1086, opened 2023-07-03
fix(RHIF-264):  adopt useChrome updateDocumentTitle hook

adonispuente:
# Description
A partial cause of https://issues.redhat.com/browse/RHIF-207 is that the document title isnt updated properly. This PR introduces the chrome hook.
Associated Jira ticket: # (RHIF-266)

# How to test the PR
Navigate to all pages and verify that the document changes appropriately 

# Checklist:

### DIFF
--- a/src/SmartComponents/Packages/Packages.js
+++ b/src/SmartComponents/Packages/Packages.js
@@ -1,5 +1,5 @@
 import { Main } from '@redhat-cloud-services/frontend-components/Main';
-import React from 'react';
+import React, { useEffect } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import messages from '../../Messages';
 import packagesListStatusFilter from '../../PresentationalComponents/Filters/PackagesListStatusFilter';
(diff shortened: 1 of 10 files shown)


src/SmartComponents/Packages/Packages.js
```diff
@@ -9,20 +9,24 @@ import TableView from '../../PresentationalComponents/TableView/TableView';
...
 
 const Packages = () => {
     const dispatch = useDispatch();
     const [firstMount, setFirstMount] = React.useState(true);
     const history = useHistory();
-    const pageTitle = 'Packages';
-    setPageTitle(pageTitle);
+    const chrome = useChrome();
+    useEffect(()=>{
+        chrome.updateDocumentTitle(`Packages ${DEFAULT_PATCH_TITLE}`);
```
leSamo, line 27: `intl` should be used instead here instead of `Packages `

leSamo (changes requested):
Thank you, I have only two small nitpicks -- one is in a separate comment, and the other one is that the dependency arrays of `useEffect` hook are a bit inconsistent, sometimes they include `intl`, sometimes not, I think it doesn't need to include `chrome` nor `intl`. But if it's consistent across all `useEffect` hooks related to titles it's okay with me.

leSamo (approved):

PR #1088, opened 2023-07-11
chore: Align semantic release config

gkarat:
This should remove semantic-release commits and stop triggering CI twice for them. Copy-pasted from Inventory


mkholjuraev:
Merging this PR to disable further semantic release commits. Hoping this will get resolved:

image

PR #1094, opened 2023-07-18
Set page title from Template detail page

leSamo:
Previously the Template name was not reflected in the title.


AsToNlele (approved):
LGTM, thank you! :smile:

PR #1107, opened 2023-08-04
Disable row kebab on Advisory Systems table when system is Applicable

leSamo:
https://issues.redhat.com/browse/SPM-2143

If system status is Installable the kebab should be enabled.


bastilian (approved):
LGTM! Thank you @leSamo!

PR #1108, opened 2023-08-08
Update @patternfly/react-core

leSamo:
Also fixed a bug where opening DatePicker (in Template Wizard) would crash the app.


AsToNlele (approved):
LGTM, thank you! :smile:

PR #1109, opened 2023-08-17
fix(RHINENG-1260): OS filter works after refresh

AsToNlele:
# Description

Associated Jira ticket: [RHINENG-1260](https://issues.redhat.com/browse/RHINENG-1260)

When reloading the page the `currentValue` is loaded from the URL and is an `array`. But when you apply an OS filter through the UI it is an `string`. I've added a check for an array so the page doesn't crash.

# How to test the PR

In the systems table apply an OS filter and refresh, the load should be successful.

# Checklist:

### DIFF
--- a/src/PresentationalComponents/Filters/OsVersionFilter.js
+++ b/src/PresentationalComponents/Filters/OsVersionFilter.js
@@ -35,7 +35,7 @@ const useOsVersionFilter = (currentFilter = '', apply) => {
         }
     }, [versionsLoaded]);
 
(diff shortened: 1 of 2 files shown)


bastilian (approved):
LGTM! Thank you @AsToNlele!

Side note: The comment does feel a bit redundant. :) It states the already obvious, but doesn't explain why it should not split in case it is an array, or why it can be an array in some cases. However, these things could also be described in a test and it would not need any documentation in the code, but the tests serve as documentation.

AsToNlele:
Thank you! Removed the comment and included some tests :smile:

PR #1113, opened 2023-08-28
Fix System list installable advisories empty state text

leSamo:
https://issues.redhat.com/browse/SPM-2085

On the System list page the "Installable advisories" column used to say "No applicable advisories" instead of "No installable advisories" when the system had no advisories.

## Before:
Screenshot from 2023-08-28 10-08-45

## After:
Screenshot from 2023-08-28 10-09-07


mkholjuraev (approved):
LGTM!

PR #1114, opened 2023-08-28
feat(ESSNTL-3716): custom filter and column for inv groups

Fewwy:
# Description

https://issues.redhat.com/browse/ESSNTL-3716

1)groups filter
2)groups column
3)filters are added to the url in format "group_name%5D=in%NAMEOFTHEGROUP", if you filter by several groups they should be separated by comma.
4) Use unleash feature flag to check if the filter and column are hidden if the feature flag is off

# How to test the PR
check that filter works and column is populated

# After the change
image

image


src/Utilities/useFeatureFlag.js
```diff
@@ -0,0 +1,7 @@
+import { useFlag, useFlagsStatus } from '@unleash/proxy-client-react';
+
+export default (flag) => {
```
mkholjuraev, line 3: There is a [hook](https://github.com/RedHatInsights/patchman-ui/blob/6c16d6d18b1c5fb8738b86b3958a8a18008a86da/src/Utilities/Hooks.js#L281C14-L281C28) for the same purpose, but using isBeta flag for some obsolete reason. Let's make a real use of it.

Fewwy: Yep I'm removing it and moving the hook to the separate file because it's easier to find and import.

adonispuente (commented):
When adding the group filter, when you go back in history, it does not update the filters. so if you
Add a group to filter through, url is updated
If I go back, the url is updated but the filters are persistent
Is this behavior alright/intended?

Also, if I open up the drop down for the groups, If i add two groups and then remove them, I am greeted with no systems/groups in the table when I should see the table filled as it is on page load. It works with reset filters, but not when unclicking items in the dropdown

src/PresentationalComponents/Filters/GroupsFilter.js
```diff
@@ -0,0 +1,42 @@
+/* eslint-disable camelcase */
+import { useEffect, useState } from 'react';
+import { getSystemsGroups } from '../../Utilities/api';
+
+const useGroupsFilter = (currentFilter = '', apply) => {
+    const [groups, setSystemsGroups] = useState([]);
+
+    useEffect(() => {
+        /* explicitly request OS versions from API */
```
mkholjuraev, line 9: Seems this comment needs to be changed ?

src/Utilities/SystemsHelpers.js
```diff
@@ -9,8 +9,9 @@ import { packageSystemsColumns } from '../SmartComponents/Systems/SystemsListAss
 import { defaultCompoundSortValues } from './constants';
 import { patchSetDetailColumns } from '../SmartComponents/PatchSetDetail/PatchSetDetailAssets';
 import { InsightsLink } from '@redhat-cloud-services/frontend-components/InsightsLink';
+import isEmpty from 'lodash/isEmpty';
 
-export const buildFilterConfig = (search, filter, apply, osFilterConfig) => ({
+export const buildFilterConfig = (search, filter, apply, osFilterConfig, groupsFilterConfig, featureFlag) => ({
```
mkholjuraev, line 14: Neatpick: it is better to name the feature flag explicitly as it belongs to the group feature only. For example, **isGroupsEnabled**

src/Utilities/SystemsHelpers.js
```diff
@@ -49,16 +51,23 @@ export const buildActiveFiltersConfig = (filter, search, deleteFilters) => ({
...
+    const tagsColumn = defaultColumns.filter(({ key }) => key === 'tags');
 
-    let name = defaultColumns.filter(({ key }) => key === 'display_name');
-    let tag = defaultColumns.filter(({ key }) => key === 'tags');
-
-    name = [{
-        ...name[0],
-        renderFunc: (displayName, id) => <InsightsLink to={`/systems/${id}`}>{displayName}</InsightsLink>
-    }];
+    groupColumn[0]?.renderFunc ? groupColumn[0].renderFunc = (data, ...value) => {
```
mkholjuraev, line 58: According to my latest review of Georgii's [PR](https://github.com/RedHatInsights/insights-advisor-frontend/pull/1166/files#diff-fa905a9ffbddeb89214abf204bfdbdf2b254bd70fb83702b7d22c3adfeb5a7c6L16) for advisor, we do not have to tweak here anymore.

mkholjuraev (commented):
A few nitpicks with the issue @adonispuente reported. Then we are good to go!

src/PresentationalComponents/Filters/GroupsFilter.js
```diff
@@ -0,0 +1,42 @@
...
+    return {
+        label: 'Group',
+        value: 'group-host-filter',
+        type: 'checkbox',
+        filterValues: {
+            onChange: (event, value) => {
+                filterByGroup(value);
+            },
+            value: currentValue,
+            //TODO: replace with map?
```
gkarat, line 33: Yeah, I think it's better function here

src/Utilities/SystemsHelpers.js
Fewwy: True! now it works like a charm without these tricks!

src/PresentationalComponents/Filters/GroupsFilter.js
Fewwy: Yep, removed

src/PresentationalComponents/Filters/GroupsFilter.js
```diff
@@ -0,0 +1,42 @@
+/* eslint-disable camelcase */
+import { useEffect, useState } from 'react';
+import { getSystemsGroups } from '../../Utilities/api';
+
+const useGroupsFilter = (currentFilter = '', apply) => {
+    const [groups, setSystemsGroups] = useState([]);
+
+    useEffect(() => {
+        /* explicitly request OS versions from API */
+        getSystemsGroups().then(({ results }) => {
```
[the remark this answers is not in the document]
Fewwy: Yep, done in my update

src/PresentationalComponents/Filters/GroupsFilter.js
Fewwy: Great, refactored

adonispuente (dismissed):
great job LGTM!

src/Utilities/SystemsHelpers.js
```diff
@@ -9,24 +9,26 @@ import { packageSystemsColumns } from '../SmartComponents/Systems/SystemsListAss
...
-            )
-        ),
-        staleFilter(apply, filter),
-        systemsUpdatableFilter(apply, filter),
-        ...osFilterConfig,
-        ...(featureFlag ? [groupsFilterConfig] : [])
-    ]
-});
+export const buildFilterConfig = (search, filter, apply, osFilterConfig, groupsFilterConfig, isGroupsEnabled) => {
+    console.log(groupsFilterConfig, 'groupsFilterConfig');
```
gkarat, line 14: Need to remove it

gkarat (approved):

PR #1123, opened 2023-09-21
chore(test): enzyme to rtl migration

Fewwy:
This PR contains only enzyme to rtl changes


src/SmartComponents/AdvisorySystems/AdvisorySystem.test.js
```diff
@@ -8,9 +7,12 @@ import { initMocks } from '../../Utilities/unitTestingUtilities.js';
 import AdvisorySystems from './AdvisorySystems';
 import { BrowserRouter as Router } from 'react-router-dom';
 import NoRegisteredSystems from '../../PresentationalComponents/Snippets/NoRegisteredSystems';
+import ShallowRenderer from 'react-test-renderer/shallow';
```
bastilian, line 10: Why does this test need to use `react-test-renderer`?

src/SmartComponents/PackageDetail/PackageDetail.test.js
```diff
@@ -60,10 +56,18 @@ afterEach(() => {
...
+        const { asFragment } = render(
+            <Provider store={store}>
+                <Router><PackageDetail /></Router>
+            </Provider>
+        );
+        expect(asFragment()).toMatchSnapshot();
     });
 
     it('should display Unavailable component on error', () => {
+        mount(<Provider store={store}>
```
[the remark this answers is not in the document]
Fewwy: Removed it and put it back into beforeEach

src/SmartComponents/AdvisorySystems/AdvisorySystem.test.js
Fewwy: You are correct, asFragment is better in this case, catches more details in the snapshots

package.json
```diff
@@ -95,12 +98,13 @@
...
     "husky": "^8.0.3",
     "identity-obj-proxy": "^3.0.0",
-    "jest": "^26.6.3",
+    "jest": "^29.6.4",
     "node-sass": "^7.0.3",
     "npm-run-all": "^4.1.5",
     "postcss": "^8.4.21",
     "prop-types": "^15.8.1",
     "react-intl": "^5.25.1",
+    "react-test-renderer": "^17.0.2",
```
bastilian, line 107: Let's try to not introduce `react-test-renderer` and use an RTL alternative test to using `ShallowRenderer`.

src/SmartComponents/AdvisorySystems/AdvisorySystem.test.js
```diff
@@ -89,7 +89,12 @@ afterEach(() => {
 describe('AdvisorySystems.js', () => {
 
     it('Should match the snapshots and dispatch FETCH_AFFECTED_SYSTEMS only once', () => {
```
bastilian, line 91: This test here and the snapshot is something we should replace with something other than `asFragment`.

The old test was already kind of useless. However, the title here does suggest a better way to test here.

Instead of trying to render, we could verify that it dispatches the right action and fetch (and only once) and to make it even better we should verify it gets dispatched for the correct advisory name. 

We could also try to not mock too much and let it all run through and verify via an rtl query that a element with the correct advisory name is present on screen.

src/SmartComponents/PackageDetail/PackageDetail.test.js
```diff
@@ -60,7 +61,12 @@ afterEach(() => {
 
 describe('PackageDetail.js', () => {
     it('should match the snapshot', ()  => {
-        expect(toJson(wrapper)).toMatchSnapshot();
+        const { asFragment } = render(
```
bastilian, line 64: I think we should change this test and instead of taking a snapshot, verify via a RTL query that a title for the loaded package (or some other uniq property/element) exists on screen.

src/SmartComponents/Packages/Packages.test.js
```diff
@@ -64,7 +64,12 @@ afterEach(() => {
 
 describe('Packages.js', () => {
     it('should match the snapshot', () => {
-        expect(toJson(wrapper)).toMatchSnapshot();
+        const { asFragment } = render(
```
bastilian, line 67: This test is useless as well. It doesn't even render any packages[0].
Let's drop this one, and maybe extend the test below and verify a list of pacakges is actually shown after it dispatched fetching.

[0] https://github.com/RedHatInsights/patchman-ui/pull/1123/files#diff-c738ba934bfc4b8dba39e7de8858f7aeaa0f81987ec0e3e5a62a1440289c9b80R261

package.json
Fewwy: Got it! Thanks, I'm gonna do that :)

Fewwy: Issue with TableView.test.js
if I replace the shallow test with asFragment - it fails due to useDispatch not being mocked. If I mock it, it fails with the following error:

 Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: object.

    Check the render method of `AsyncRemediationButton`.

Google says that's an incorrect import issue. I tried the proposed solutions (add curly braces) and it's not the "fix"

src/SmartComponents/AdvisorySystems/AdvisorySystem.test.js
Fewwy: I agree, I tried to replace the asFragment with something that testing the actual inventory table that is rendered. 
I tried the following
```
render (our component)
and then use
screen.debug() to log the DOM.
```
This is what I get in the terminal, why the table is not rendering? Do you have any idea?
image

src/SmartComponents/Packages/Packages.test.js
Fewwy: Added 2 tests to check the packages names and summary

src/SmartComponents/PackageDetail/PackageDetail.test.js
Fewwy: Added 2 tests that checks the component text

package.json
bastilian: The mock in `setupTests.js` is not correct. it should be:

```js
 jest.mock(
    '@redhat-cloud-services/frontend-components/AsyncComponent',
    () => ({
        __esModule: true,
        default: (props) => <div {...props}>AsyncComponent</div>,
    })
);
```

src/SmartComponents/AdvisorySystems/AdvisorySystem.test.js
bastilian: Are you providing the necessary data for it to render the correct state?

src/SmartComponents/AdvisorySystems/AdvisorySystem.test.js
```diff
@@ -8,8 +7,11 @@ import { initMocks } from '../../Utilities/unitTestingUtilities.js';
 import AdvisorySystems from './AdvisorySystems';
 import { BrowserRouter as Router } from 'react-router-dom';
 import NoRegisteredSystems from '../../PresentationalComponents/Snippets/NoRegisteredSystems';
+import { render, act as rtlAct } from '@testing-library/react';
```
bastilian, line 10: You can remove the other outdated not to be used anymore act, and not need to rename it.

src/SmartComponents/AdvisorySystems/AdvisorySystem.test.js
```diff
@@ -8,8 +7,11 @@ import { initMocks } from '../../Utilities/unitTestingUtilities.js';
...
 
 initMocks();
+//removing act warnings from the console
+const promise = Promise.resolve();
```
bastilian, line 14: no.

src/SmartComponents/PackageDetail/PackageDetail.test.js
```diff
@@ -41,6 +42,7 @@ const initStore = (state) => {
     return mockStore({ PackageDetailStore: state });
 };
 
+// eslint-disable-next-line no-unused-vars
```
bastilian, line 45: Please, don't do this. Correct what is not right and don't ignore it.

src/SmartComponents/AdvisoryDetail/CveModal.test.js
```diff
@@ -51,7 +51,7 @@ beforeEach(() => {
     createCvesRows.mockImplementation(() => readyCveRows);
 
     wrapper = mount(<Provider store={store}>
-        <Router><CvesModal cveIds={['testCveID']}  /></Router>
+        <Router><CvesModal /></Router>
```
bastilian, line 54: Why is the `cveIds` prop here removed?

src/PresentationalComponents/TableView/TableView.test.js
```diff
@@ -1,6 +1,6 @@
 import TableView from './TableView';
-import toJson from 'enzyme-to-json';
 import NoRegisteredSystems from '../../PresentationalComponents/Snippets/NoRegisteredSystems';
+import ShallowRenderer from 'react-test-renderer/shallow';
```
bastilian, line 3: Replace with RTL tests.

src/PresentationalComponents/Header/Header.test.js
```diff
@@ -100,20 +101,20 @@ describe('Header component', () => {
...
             ]
         };
-        const wrapper = withLink(<Header {...header} />);
-        expect(toJson(wrapper)).toMatchSnapshot();
+        const { asFragment } = withLink(<Header {...header} />);
+        expect(asFragment()).toMatchSnapshot();
     });
-
-    it('should render with tabs only, click a tab', () => {
+    //NOTE: Should be rewritten because of wrapper changes
```
bastilian, line 107: Because of what changes?

src/PresentationalComponents/Header/Header.test.js
```diff
@@ -1,32 +1,33 @@
 /* eslint-disable */
-import toJson from 'enzyme-to-json';
 import React from 'react';
-import { BrowserRouter as Router } from 'react-router-dom';
+import { BrowserRouter as Router} from 'react-router-dom';
 import Header from './Header';
+import { render } from '@testing-library/react';
+
 
 const withLink = Component => {
```
bastilian, line 8: We should rename this to something more fitting like `Wrapper`. Additionally, i'd not call the `render` here but separately in each test case.

package.json
Fewwy: Great catch! Thanks <3

src/SmartComponents/AdvisorySystems/AdvisorySystem.test.js
Fewwy: Ok

src/SmartComponents/PackageDetail/PackageDetail.test.js
Fewwy: You are right, removed all of these exceptions

src/PresentationalComponents/Header/Header.test.js
Fewwy: by renaming this to something more fitting - you mean to create a function for wrapping the components like here? 
https://testing-library.com/docs/react-testing-library/setup/

src/PresentationalComponents/Header/Header.test.js
Fewwy: I formulated that incorrectly, due to how RTL works we shouldn't test the history changes from the inside the component. If we want to test it we should do it separately https://stackoverflow.com/questions/70313688/how-i-could-test-location-with-memoryrouter-on-react-router-dom-v6

src/SmartComponents/AdvisoryDetail/CveModal.test.js
Fewwy: Good catch!

src/PresentationalComponents/Header/Header.test.js
bastilian: I mean this should be something named similar to this wrapper here: https://github.com/RedHatInsights/insights-inventory-frontend/pull/2033/files#diff-a24acf51249c422f4ce15987438346f5ef701213be1e12c72125dcf53832deccR13

Also note the move of `render` out of a function like this.

src/PresentationalComponents/Header/Header.test.js
```diff
@@ -100,20 +101,20 @@ describe('Header component', () => {
...
-        const wrapper = withLink(<Header {...header} />);
-        expect(toJson(wrapper)).toMatchSnapshot();
+        const { asFragment } = withLink(<Header {...header} />);
+        expect(asFragment()).toMatchSnapshot();
     });
-
-    it('should render with tabs only, click a tab', () => {
+    //NOTE: Should be rewritten because of how RTL testing approaches testing the router
+    //something like this could be ok https://stackoverflow.com/questions/70313688/how-i-could-test-location-with-memoryrouter-on-react-router-dom-v6
+    /* it('should render with tabs only, click a tab', () => {
```
bastilian, line 109: Don't comment out tests. use `.skip`.

PR #1132, opened 2023-10-20
chore(deps): setting up rtl and updating packages for future pr

Fewwy:
# Description
I am separating my work on the RTL into smaller pieces so it would be easier to review and merge


config/rtlwrapper.js
```diff
@@ -0,0 +1,15 @@
...
+import { Provider } from 'react-redux';
+
+export const mountWithRouter = (Component) => {
+    return render(<BrowserRouter>{Component}</BrowserRouter>);
+};
+
+export const mountWithRouterAndProvider = (Component, store) => {
+    return render(
+        <Provider store={store}>
+            <BrowserRouter>{Component}</BrowserRouter>
```
mkholjuraev, line 12: Wondering if MemoryRouter would be better to use?

mkholjuraev (approved):
LGTM!

config/rtlwrapper.js
Fewwy: I tried it and it didn't work :(

PR #1133, opened 2023-10-20
Enzyme tests refactored separated by commits

Fewwy:
This is the same as https://github.com/RedHatInsights/patchman-ui/pull/1123 PR but separated by commits/file changed


johnsonm325 (approved):
Approved. But I do have one question. It looks like the snapshots now remove the names of the child components and replaces them with pure html like `<h5>` and `<div>` rather than `<Breadcrumb>` and the like. Is this the kind of output we expect to see from RTL?

Fewwy:
@johnsonm325 As far as I understand - yes

PR #1136, opened 2023-10-26
chore: Update npm run start scripts and documentation

gkarat:
no jira.

This is a quick PR that aims to make npm run start* scripts consistent across Insights frontend applications.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,8 +28,6 @@ There is [a comprehensive quick start guide in the Storybook Documentation](http
 
 Note: You will need to set up the Insights environment if you want to develop with the starter app due to the consumption of the chroming service as well as setting up your global/app navigation through the API.
 
-
-
 ## Run the app for the first time
 
 1. ```git clone git@github.com:RedHatInsights/patchman-ui.git```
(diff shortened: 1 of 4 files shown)


gkarat:
Relates to https://github.com/RedHatInsights/vulnerability-ui/pull/2013

gkarat:
@mkholjuraev, could you please help me checking if dev.webpack.config.js contains all the necessary parameters? For now, I put only those that are common across all apps. I am not sure about some extra ones that Patch added like useCloud or useFileHash. Since you are more experienced with Patch, you might have more insights if these parameters are used at all and if we need to keep them further

mkholjuraev (commented):
The app runs pretty well. According to team's decision, we can keep or remove following params:

1. useFileHash: this is needed to turn on caching in dev env and platform advices to do so. Ref: [consuming-module](http://front-end-docs-insights.apps.ocp4.prod.psi.redhat.com/fed/consuming-module) 
2. debug: as this is dev env, do we want to have debug mode enabled?
3. proxyVerbose: this is intended for the output of what the proxy is doing when we run the app. I am not sure if we need it (Attached image)

PR #1141, opened 2023-11-03
fix(InventoryTable): RHINENG-2912 - Remove custom OS filter handling

bastilian:
# Description

Associated Jira ticket: #RHINENG-2912

Removes the custom handling for the OS filter for InventoryTables

# Before the change

https://github.com/RedHatInsights/patchman-ui/assets/7757/2c301da9-6150-41df-8690-42022b7de260

# After the change

https://github.com/RedHatInsights/patchman-ui/assets/7757/1a348481-62b6-4c56-8d4e-15933b5590b8

### DIFF
--- a/src/SmartComponents/AdvisorySystems/AdvisorySystems.js
+++ b/src/SmartComponents/AdvisorySystems/AdvisorySystems.js
@@ -6,7 +6,6 @@ import { shallowEqual, useDispatch, useSelector, useStore } from 'react-redux';
 import { combineReducers } from 'redux';
 import messages from '../../Messages';
 import searchFilter from '../../PresentationalComponents/Filters/SearchFilter';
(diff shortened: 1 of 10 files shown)


mkholjuraev (changes requested):
Codewise looks clean. While testing, I found out that appliying rhel 8.8 does not filter the table correctly. I see that there are rhel 8.8 systems, but applying the filter results in an empty result.  It seems filtering syntax got affected, it has incorrect value : **in:in:RHEL 8.8**

bastilian:
@mkholjuraev Thank you for reviewing and testing it! The filtering should now work as expected again.

mkholjuraev (approved):
LGTM! I see that the new function is covered with tests. We can enhance it more by writing tests that ensure that actvie OS filter indeed goes to the API request as expected.

PR #1144, opened 2023-11-24
Disable template edit actions when missing permission

leSamo:
Associated Jira ticket: [RHINENG-3143](https://issues.redhat.com/browse/RHINENG-3143)

When user is missing permissions to edit templates:
- Disable `Apply to systems` button in `NoAppliedSystems` empty state in Template detail page
- Disable edit and delete template buttons in Template detail page header dropdown
- Disable template add and remove template buttons is System detail header dropdown
- Disable add and remove template table actions (both bulk and row) in System list table

I tried to add tooltip why the button is disabled where available, but due to the limitations of external components it was not always possible


bastilian (approved):
LGTM! Thank you @leSamo!

PR #1152, opened 2023-12-12
fix(OperatingSystemFilter): RHINENG-10345 - Fix OS inventory filter conversion

bastilian:
This fixes issues with syncing the state and URL of filters in the Systems table.

# Description

# How to test the PR

# Before the change

# After the change

# Dependent work link

# Checklist:

### DIFF
--- a/src/SmartComponents/Systems/SystemsTable.js
+++ b/src/SmartComponents/Systems/SystemsTable.js
@@ -60,7 +60,23 @@ const SystemsTable = ({
     ]);
 
     const { systemProfile, selectedTags,
-        filter, search, page, perPage, sort } = queryParams;
+        filter: queryParamsFilter, search, page, perPage, sort
+    } = queryParams;
+    const { os: operatingSystemFilter, ...filter } = queryParamsFilter || {};
+    const osFilter = operatingSystemFilter && [{
+        osFilter: operatingSystemFilter.reduce((osFilter, os) => {
+            const [osName, osVersion] = os.split(' ');
+            const [major] = osVersion.split('.');
+
+            return {
+                ...osFilter,
+                [`${osName}-${major}`]: {
+                    ...osFilter[`${osName}-${major}`] || {},
+                    [`${osName}-${major}-${osVersion}`]: true
+                }
+            };
+        }, {})
+    }];
 
     const applyMetadata = (metadata) => {
         dispatch(changeSystemsMetadata(metadata));
@@ -143,6 +159,9 @@ const SystemsTable = ({
             columns={(defaultColumns) => systemsColumnsMerger(defaultColumns, systemsListColumns)}
             showTags
             customFilters={{
+                ...operatingSystemFilter ? {
+                    filters: [...(osFilter || [])]
+                } : {},
                 patchParams: {
                     search,
                     filter,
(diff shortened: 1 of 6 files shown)


src/SmartComponents/Systems/SystemsTable.js
```diff
@@ -60,7 +60,23 @@ const SystemsTable = ({
...
+    const { os: operatingSystemFilter, ...filter } = queryParamsFilter || {};
+    const osFilter = operatingSystemFilter && [{
+        osFilter: operatingSystemFilter.reduce((osFilter, os) => {
+            const [osName, osVersion] = os.split(' ');
+            const [major] = osVersion.split('.');
+
+            return {
+                ...osFilter,
+                [`${osName}-${major}`]: {
+                    ...osFilter[`${osName}-${major}`] || {},
```
gkarat, line 74: nit-pick:
```suggestion
                    ...osFilter[`${osName}-${major}`],
```

src/SmartComponents/Systems/SystemsTable.js
```diff
@@ -143,6 +159,9 @@ const SystemsTable = ({
             columns={(defaultColumns) => systemsColumnsMerger(defaultColumns, systemsListColumns)}
             showTags
             customFilters={{
+                ...operatingSystemFilter ? {
+                    filters: [...(osFilter || [])]
+                } : {},
```
gkarat, line 164: I think this is something we could cover with tests: there is SystemTable.test.js available where we could add a scenario where query parameters contain `filter[os]`. And given that InventoryTable receives customFilters, we can assert on what the mocked component was called with and make sure osFilter is passed in the expected format. WDYT?

gkarat (commented):
Looks good! I have tried to test it, couldn't find issues and the OS filter now operates as expected. There is one thing I would like to ask for in terms of tests before we can merge...

src/SmartComponents/Systems/SystemsTable.js
bastilian: Not falling back to `{}` could cause an error, because it the major version could be missing.

bastilian: true. I confused it with `[ ...undefined ]`

src/SmartComponents/Systems/SystemsTable.js
```diff
@@ -60,7 +60,23 @@ const SystemsTable = ({
     ]);
 
     const { systemProfile, selectedTags,
-        filter, search, page, perPage, sort } = queryParams;
+        filter: queryParamsFilter, search, page, perPage, sort
+    } = queryParams;
+    const { os: operatingSystemFilter, ...filter } = queryParamsFilter || {};
+    const osFilter = operatingSystemFilter && [{
```
mkholjuraev, line 66: Can we extract this into some utility function? The os filter exists in Advisory systems table as well. Probably we need to make same changes there as welll

bastilian: AH! I'll take a look!

bastilian: There we don't need the change, because the table is not querying against the Patch API, but the inventory API and the filter comes directly from there.

bastilian: oh, I'm confused. ok looking more. before I say more. :D

mkholjuraev:
@bastilian can you please have a look at this PR. The OS filter is broken in both stage and prod and we are receiving more and more complaints unfortunately.

mkholjuraev (approved):
LGTM!

PR #1153, opened 2023-12-12
fix(SystemsTable): RHINENG-2396 - Prevent state & URL updates when not mounted

bastilian:
# Description

Associated Jira ticket: #RHINENG-2396 

This prevents patch from updating its state and the URL parameters if the component/hook got unmounted while it was fetching systems from the API

# How to test the PR

1) Navigate to the Patch -> Systems page
2) As soon as the page loads and the InventoryTable appears in the loading state navigate away to a different page.

# Before the change

Without preventing the state/URL updates Patch will reappear again once the fetching of systems completed.

# After the change

The patch systems page does not reappear.

# Checklist:

### DIFF
--- a/src/Utilities/Hooks.js
+++ b/src/Utilities/Hooks.js
@@ -1,4 +1,4 @@
-import React, { useCallback } from 'react';
+import React, { useCallback, useEffect, useRef } from 'react';
 import { useLocation, useNavigate } from 'react-router-dom';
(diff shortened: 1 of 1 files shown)


mkholjuraev (approved):
LGTM! Works as expected.  @bastilian thank you!

For my understanding, what lead us to chose refs over AbortController?

PR #1161, opened 2024-02-07
chore(RHINENG-5923): Update zero state check

adonispuente:
This updates zero state in malware to use the updated version of the fed mod, and to make the call for ONLY RHEL systems, as opposed to ALL systems


AsToNlele:
I guess prod has the old zerostate, so will do review after stage back is up

AsToNlele (approved):
LGTM, good job :smile:

PR #1162, opened 2024-02-12
fix: Remove template name regex restriction

leSamo:
https://redhat-internal.slack.com/archives/C02T90PLJ5V/p1707755885806079


gkarat (approved):
LGTM, thanks @leSamo

PR #1164, opened 2024-02-22
chore(RHINENG-7975): RTL migration

mkholjuraev:
# Description

Associated Jira ticket: # [(issue)](https://issues.redhat.com/browse/RHINENG-7975)

- cleans up the tests by removing unnecessarily mocked redux selectors, components, functions
- rewrites enzyme tests to RTL
- tests for inventory tables are commented out, for now, to unblock the migration of PF5 and react. Also, we need to find a meaningful way of testing fed-modules.

# How to test the PR

# Before the change

# After the change

# Dependent work link

# Checklist:


src/PresentationalComponents/Snippets/ErrorHandler.js
```diff
@@ -49,8 +49,8 @@ const ErrorHandler = ({ code, ErrorState, EmptyState, metadata = {} }) => {
             return <Unavailable />;
 
         default:
-            return ErrorState && <ErrorState />
-                || EmptyState && <EmptyState />
+            return ErrorState && ErrorState
```
bastilian, line 52: 👀 This looks odd. Shouldn't `ErrorState || EmptyState || ...` be enough?

src/SmartComponents/AdvisoryDetail/CveModal.test.js
```diff
@@ -29,107 +22,80 @@ jest.mock('../../Utilities/api', () => ({
...
-});
-
-afterEach(() => {
-    useSelector.mockClear();
+    render(<ComponentWithContext renderOptions={{ store }}>
+        <CvesModal />
+    </ComponentWithContext>);
 });
 
+//TODO: convert disabled tests to RTL after react&paterrnfly migration
```
bastilian, line 39: What are the issues with converting them before?

src/SmartComponents/AdvisorySystems/AdvisorySystem.test.js
```diff
@@ -49,150 +38,128 @@ const mockState = {
...
-
-afterEach(() => {
-    useSelector.mockClear();
-    store.clearActions();
+    render(<ComponentWithContext renderOptions={{ store }}>
+        <AdvisorySystems advisoryName={'RHSA-2020:2755'} />
+    </ComponentWithContext>);
 });
 
+//TODO: find a meaningful way of testing InventoryTable fed module
```
bastilian, line 52: I think there won't be a good way to do this. We can with a better mock maybe verify that some props are properly passed in and we might be able to mimic the behaviour of the InventoryTable, but this might be more headache in the long run, especially the last part. 

We should focus more on making sure that everything around works as expected. For example, rather than verifying the `getEntities` by having it called by the `InventoryTable` and expect it to render the systems in the table, call it directly and verify its output matches what the inventory would expect. Or rather than verifying the `onSelect` handler by wanting to click in the table on the checkboxes only verify the `onSelect` handler directly.

src/SmartComponents/AdvisoryDetail/CveModal.test.js
mkholjuraev: This modal does not get used in patch according to the pendo. I wanted to unblock the migrations for Patch by converting essential tests in the first place. I will start working on this and deliver it as soon as possible

src/SmartComponents/AdvisorySystems/AdvisorySystem.test.js
mkholjuraev: I agree. Actually, I have started on creating such a mocked component that just triggers events passed as a prop. I am going to create it in a separate PR as I said earlier.

bastilian: @mkholjuraev To elaborate more. I tried this[0]. It only got as far as calling the `getEntities`, because implementing everything else would have required to pretty much reimplement the InventoryTable.

[0] https://github.com/RedHatInsights/compliance-frontend/blob/master/src/SmartComponents/SystemsTable/__mocks__/InventoryTableMock.js#L9

src/SmartComponents/Advisories/Advisories.test.js
```diff
@@ -55,152 +59,66 @@ const mockState = { ...storeListDefaults,
...
-
-afterEach(() => {
-    useSelector.mockClear();
+    render(<ComponentWithContext renderOptions={{ store }}>
+        <Advisories/>
+    </ComponentWithContext>);
 });
 
 describe('Advisories.js', () => {
+    testExport(exportAdvisoriesCSV, exportAdvisoriesJSON, 'advisories');
```
gkarat, line 75: nit-pick: even though the test is shared between two different test files, I'd better have it separately duplicated in both of them. Reading this/traveling to another file module/understanding props just increases the cognitive load and makes the code less readable. The benefits of having DRY code are also not significant. IMHO

src/SmartComponents/Advisories/Advisories.test.js
```diff
@@ -55,152 +59,66 @@ const mockState = { ...storeListDefaults,
...
-        const onSelect = wrapper.find('TableView').props().onSelect;
-        onSelect('page');
-        const dispatchedActions = store.getActions();
-        expect(dispatchedActions[2].type).toEqual('SELECT_ADVISORY_ROW');
-        expect(dispatchedActions[2].payload[0]).toEqual({
-            id: 'RHEA-2020:2743',
-            selected: 'RHEA-2020:2743'
-        });
-    });
+    it('should handle remediation', () => {
```
gkarat, line 103: nit-pick: technically, the test only asserts if the button is rendered with the right props. There is no scenario that would actually check remediation handling

src/SmartComponents/Advisories/Advisories.test.js
```diff
@@ -55,152 +59,66 @@ const mockState = { ...storeListDefaults,
...
+        const tempStore = initStore(rejectedState);
+        const { container } = render(<ComponentWithContext renderOptions={{ store: tempStore }}>
+            <Advisories/>
+        </ComponentWithContext>);
 
-        expect(fetchIDs).toHaveBeenCalledWith(
-            '/ids/advisories',
-            { limit: -1, offset: 0, page: 1, page_size: 20 }
+        await waitFor(
+            () => expect(queryByText(container, 'This page is temporarily unavailable')).toBeVisible()
```
gkarat, line 99: Avoid using container. You can just use
```suggestion
            () => expect(screen.queryByText('This page is temporarily unavailable')).toBeVisible()
```
here

src/SmartComponents/AdvisoryDetail/CveModal.test.js
```diff
@@ -29,107 +22,80 @@ jest.mock('../../Utilities/api', () => ({
...
     });
+    // it('should set rows to undefined to close the modal', () => {
+    //     screen.debug(undefined, 30000);
+    //     // const handleClose = wrapper.find('Modal').props().onClose;
+    //     // handleClose();
+    //     // wrapper.update();
+    //     // expect(wrapper.find('TableView').exists()).toBeFalsy();
+    // });
+
+    // it('should handle page change', () => {
```
gkarat, line 52: Please, use `skip` instead so that the number of skipped tests is reported with each run (= we have a better reminder of some tests being omitted temporarily)

src/SmartComponents/AdvisorySystems/AdvisorySystem.test.js
```diff
@@ -49,150 +38,128 @@ const mockState = {
...
-        const tempStore = initStore(notFoundState);
-        const tempWrapper = mount(<Provider store={tempStore}>
-            <Router><AdvisorySystems /></Router>
-        </Provider>);
-
-        expect(tempWrapper.find(NoRegisteredSystems)).toBeTruthy();
+    it('Should render inventory table', () => {
+        expect(screen.getByTestId('inventory-mock-component')).toBeVisible();
     });
+    // it('Should display error page when status is rejected', () => {
```
gkarat, line 57: The same comment regarding using `skip`

src/SmartComponents/Modals/UnassignSystemsModal.test.js
```diff
@@ -62,35 +54,25 @@ describe('UnassignSystemsModal', () => {
...
         );
 
-        await wrapper.update().find(Modal).props().actions[0].props.onClick();
-        expect(addNotification).toHaveBeenCalledWith(
-            patchSetUnassignSystemsNotifications(1).success
-        );
-        expect(unassignSystemsModalState).toEqual({ isUnassignSystemsModalOpen: false, shouldRefresh: true, systemsIDs: [] });
-        expect(unassignSystemFromPatchSet).toHaveBeenCalledWith({ inventory_ids: ['test_1'] });
-    });
+        fireEvent.click(screen.getByText('Remove'));
```
gkarat, line 57: nit-pick: Prefer userEvent here
```suggestion
        userEvent.click(screen.getByText('Remove'));
```

src/SmartComponents/Modals/UnassignSystemsModal.test.js
```diff
@@ -62,35 +54,25 @@ describe('UnassignSystemsModal', () => {
...
-                isUnassignSystemsModalOpen: false,
-                systemsIDs: ['test_1', 'test_2', 'test_3']
-            }}
-            setUnassignSystemsModalOpen={setUnassignSystemsModalOpen}
-        />
-        );
+    it('should close the modal', async () => {
+        unassignSystemsModalState.isUnassignSystemsModalOpen = true;
+        screen.debug(undefined, 30000);
+        fireEvent.click(screen.getByLabelText('Close'));
```
gkarat, line 71: nit-pick:
```suggestion
        userEvent.click(screen.getByLabelText('Close'));
```

src/SmartComponents/Modals/UnassignSystemsModal.test.js
```diff
@@ -62,35 +54,25 @@ describe('UnassignSystemsModal', () => {
...
-            unassignSystemsModalState={{
-                isUnassignSystemsModalOpen: false,
-                systemsIDs: ['test_1', 'test_2', 'test_3']
-            }}
-            setUnassignSystemsModalOpen={setUnassignSystemsModalOpen}
-        />
-        );
+    it('should close the modal', async () => {
+        unassignSystemsModalState.isUnassignSystemsModalOpen = true;
+        screen.debug(undefined, 30000);
```
gkarat, line 70: I think we no longer need that?

src/SmartComponents/PackageSystems/PackageSystems.test.js
```diff
@@ -1,23 +1,14 @@
-import { Provider, useSelector } from 'react-redux';
-import { BrowserRouter as Router } from 'react-router-dom';
+/* eslint-disable no-unused-vars */
```
gkarat, line 1: Why we need to ignore this rule here?

src/SmartComponents/SystemAdvisories/SystemAdvisories.test.js
```diff
@@ -40,153 +41,67 @@ const mockState = {
...
-        expect(dispatchedActions.filter(item => item.type === 'EXPAND_SYSTEM_ADVISORY_ROW')).toHaveLength(1);
-        expect(dispatchedActions[1].payload).toEqual({
-            rowId: 'RHSA-2020:2774',
-            value: 'testValue'
-        });
-    });
-
-    it('Should collapse a row ', () => {
-        wrapper.find('TableView').props().onCollapse(null, 0, 'testValue');
+    it('should handle remediation', async() => {
```
gkarat, line 56: nit-nit-nit-pick:
```suggestion
    it('should handle remediation', async () => {
```

Alt Text

Doesn't your linter/prettier automatically format the code for you? 🤔

src/SmartComponents/SystemAdvisories/SystemAdvisories.test.js
```diff
@@ -40,153 +41,67 @@ const mockState = {
...
-            value: 'testValue'
+        const tempStore = initStore(selectedState);
+        render(
+            <ComponentWithContext renderOptions={{ store: tempStore }}>
+                <SystemAdvisories inventoryId='test' />
+            </ComponentWithContext>
+        );
+
+        waitFor(() => {
+            expect(AsyncRemediationButton).toHaveBeenCalledWith({
```
gkarat, line 70: Again, I feel that the title of the test is not aligned with what is actually tested

src/SmartComponents/SystemAdvisories/SystemAdvisories.test.js
```diff
@@ -40,153 +41,67 @@ const mockState = {
...
-        const tempWrapper = mount(<Provider store={tempStore}>
-            <Router><SystemAdvisories inventoryId='test' /></Router>
-        </Provider>);
+        const tempStore = initStore(emptyState);
+        const { container } = render(<ComponentWithContext renderOptions={{ store: tempStore }}>
+            <SystemAdvisories inventoryId='test' />
+        </ComponentWithContext>);
 
-        expect(tempWrapper.find('SystemUpToDate')).toBeTruthy();
+        await waitFor(() => queryByText(container, 'No applicable advisories'));
```
gkarat, line 104:
```suggestion
        await waitFor(() => screen.getByText('No applicable advisories'));
```

src/SmartComponents/SystemPackages/SystemPackages.test.js
```diff
@@ -39,149 +32,123 @@ jest.mock('../../Utilities/Helpers', () => ({
...
+    it('Should provide correct remediation data', async () => {
+        const selectedState = {
+            ...mockState,
+            selectedRows: { 'test-id-0': 'test-id-0' }
+        };
 
-    it('Should dispatch FETCH_APPLICABLE_SYSTEM_PACKAGES only once on load', () => {
-        const dispatchedActions = store.getActions();
-        expect(dispatchedActions.filter(item => item.type === 'FETCH_APPLICABLE_SYSTEM_PACKAGES')).toHaveLength(1);
+        await mountComponent(selectedState);
```
gkarat, line 91:
```suggestion
        mountComponent(selectedState);
```

src/SmartComponents/SystemPackages/SystemPackages.test.js
```diff
@@ -39,149 +32,123 @@ jest.mock('../../Utilities/Helpers', () => ({
...
+        await waitFor(() => screen.getByLabelText('Patch table view'));
+
+        fireEvent.click(screen.getByText('Remediation'));
```
gkarat, line 94:
```suggestion
        userEvent.click(screen.getByText('Remediation'));
```

src/Utilities/hooks/Hooks.test.js
```diff
@@ -94,79 +79,61 @@ describe('Custom hooks tests', () => {
...
     ${{advisory_type: [2,1]}} | ${jest.fn()} | ${[{ id: "advisory_type", chips: [{id: 1}] }]} | ${{filter: {advisory_type: [2]}}}
     ${{search: "asd"}}        | ${jest.fn()} | ${[{ id: "search"}]}                           | ${{filter: {}, search: ''}}
-    `('useRemoveFilter: should return correct filter for $filter',({filter, apply, result, selected}) => {
-        let deleteFilters;
-        testHook(() => {
-            [deleteFilters] = useRemoveFilter(filter, apply);
-        });
-        deleteFilters({}, selected)
-        expect(apply).toHaveBeenCalledWith(result);
+    `('useRemoveFilter: should return correct filter for $filter',({filter, apply, finalResult, selected}) => {
```
gkarat, line 102: Wow, never seen this method

gkarat: Perhaps we could use this variant instead https://jestjs.io/docs/api#1-describeeachtablename-fn-timeout? Seems to me more readable. But it's mine subjective opinion

gkarat, line 102:
```suggestion
    `(`useRemoveFilter: should return correct filter for ${filter}`,({filter, apply, finalResult, selected}) => {
```

src/Utilities/TestingUtilities.js
```diff
@@ -0,0 +1,104 @@
...
+            expect.objectContaining({ limit: -1, offset: 0 })
+        );
+    });
+
+    it('should select rows', async () => {
+        const spy = jest.spyOn(Actions, 'selectAdvisoryRow');
+        fireEvent.click(screen.getByLabelText('Select'));
+        fireEvent.click(screen.getByText('Select page (1)'));
+
+        waitFor(() =>
```
gkarat, line 89:
```suggestion
        await waitFor(() =>
```
waitFor's should be awaited IIRC

src/Utilities/TestingUtilities.js
```diff
@@ -0,0 +1,104 @@
...
+            expect(spy).toHaveBeenCalledWith(selectedItem)
+        );
+    });
+
+    it('should unselect rows', async () => {
+        const spy = jest.spyOn(Actions, spyOnAction);
+        fireEvent.click(screen.getByLabelText('Select'));
+        fireEvent.click(screen.getByText('Select none (0)'));
+
+        waitFor(() =>
```
gkarat, line 99:
```suggestion
        await waitFor(() =>
```

src/Utilities/__snapshots__/DataMappers.test.js.snap
```diff
@@ -1,34 +1,3 @@
...
-      "title": "Access control list utilities",
-    },
-  ],
-  "disableSelection": false,
-  "id": "test-name-2-test-evra-2",
-  "key": "test-name-2-test-evra-2",
-  "selected": false,
-}
-`;
+exports[`DataMappers Should create package list 1`] = `undefined`;
```
gkarat, line 3: Shouldn't this snapshot be removed completely?

src/SmartComponents/Advisories/Advisories.test.js
mkholjuraev: I would argue that having duplicate tests for the same feature would require me to to each test file and make the same changes. That is more time-consuming and cumbersome than cognitive load :)

src/SmartComponents/Modals/UnassignSystemsModal.test.js
mkholjuraev: I do not see a huge gain in using `userEvent` when I need to only simulate the onClick event.

src/SmartComponents/AdvisoryDetail/CveModal.test.js
mkholjuraev: Good call. will do

src/SmartComponents/Modals/UnassignSystemsModal.test.js
gkarat: > user-event allows you to describe a user interaction instead of a concrete event. It adds visibility and interactability checks along the way and manipulates the DOM just like a user interaction in the browser would. It factors in that the browser e.g. wouldn't let a user click a hidden element or type in a disabled text box.
> This is [why you should use user-event](https://ph-fritsche.github.io/blog/post/why-userevent) to test interaction with your components.

src/SmartComponents/AdvisorySystems/AdvisorySystem.test.js
mkholjuraev: I think I could not explain my point, but I believe we are talking about the same thing

src/Utilities/hooks/Hooks.test.js
mkholjuraev: This is expected. The method syntax needs to have **'useRemoveFilter: should return correct filter for $filter'** to display the messages customised to each filter. Leaving as it is

src/SmartComponents/AdvisorySystems/AdvisorySystem.test.js
gkarat: Not sure I am gonna add anything more than what Sebastian already described. I am against re-implementing InventoryTable just for the sake of testing it. Because of its federated module nature, the best we can do here is to assert on props. For example, I would just check:

- if `hideFilters` has tags and operatingSystem filter set to true (= we want InventoryTable to show only these two filters)
- `columns` receives the function and its return value matches our expectations of the columns we want to see
- that `customFilters` contains the configuration of the custom filters we want to add on top of tags and operatingSystem filters
- `paginationProps` has `isDisabled` set to true if total items are 0
- with `dedicatedAction` we pass the remediation button.

These are just a half of other things we can assert on, I listed them above just for better understanding.

I am also not sure we even want to test that `getEntities` or something else is called. Mainly because this will inevitable involve adding mocked implementation which will be different from and not synced with the real InventoryTable behaviour - this doesn't make sense to test.

gkarat:
@mkholjuraev, thanks for addressing my comments! Let's finish on InventoryTable part... other than that LGTM.

mkholjuraev:
> @mkholjuraev, thanks for addressing my comments! Let's finish on InventoryTable part... other than that LGTM.

@gkarat thank you for another look. Maybe, we could merge this PR and work in parallel with you? I will continue with the tests for the Inventory pages and you can start your work on the migration.

gkarat (approved):
Agree, let's merge it. Thanks for this huge chunk of tests improved

src/SmartComponents/AdvisorySystems/AdvisorySystem.test.js
mkholjuraev: Yup, I agree. I have previously created similar tests for the ImmutableDevices table that uses the inventory-fed module. Those tests make sure that the correct filters are passed, the correct state is present if there is an error, etc.

Example: https://github.com/RedHatInsights/vulnerability-ui/blob/master/src/Components/SmartComponents/HybridInventory/ImmutableDevicesTab/ImmutableDevices.test.js

mkholjuraev:
the pr_check is failing due to vmass sync. Ignoring it for now.

PR #1167, opened 2024-03-01
chore(deps): Update axios to fix CVE

leSamo:
Fixes https://issues.redhat.com/browse/RHINENG-8635

Axios did have an [error handler refactoring](https://github.com/axios/axios/pull/3645) in [0.27.0](https://github.com/axios/axios/releases/tag/v0.27.0), therefore I had to change the accessor to `isAxiosError`.

### DIFF
--- a/src/Utilities/axiosInterceptors.js
+++ b/src/Utilities/axiosInterceptors.js
@@ -9,18 +9,16 @@ const axiosInstance = axios.create();
 export function errorInterceptor(err) {
 
     if (!axios.isCancel(err)) {
+        const { response } = { ...err };
 
-        const { response, isAxiosError } = { ...err };
-
-        if (response && isAxiosError) {
+        if (response && err.isAxiosError) {
             const { status, statusText, data } = response;
 
             if (!status) {
                 return err;
             } else {
                 const genericError = {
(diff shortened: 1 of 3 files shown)


src/Utilities/axiosInterceptors.js
```diff
@@ -9,18 +9,16 @@ const axiosInstance = axios.create();
 export function errorInterceptor(err) {
 
     if (!axios.isCancel(err)) {
+        const { response } = { ...err };
 
-        const { response, isAxiosError } = { ...err };
-
-        if (response && isAxiosError) {
+        if (response && err.isAxiosError) {
```
mkholjuraev, line 14: For my understanding, what was the issue here?

mkholjuraev (approved):
LGTM!

src/Utilities/axiosInterceptors.js
leSamo: In the new version of axios

```js
const { response, isAxiosError } = { ...err }
console.log(isAxiosError); // prints "undefined"
console.log(err.isAxiosError); // prints "true"
```

It seems like by copying the object the destructuring and direct property access behave differently.

PR #1176, opened 2024-03-14
fix(npm): audit fix modules

Fewwy:
Small update of packages running npm audit fix

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4663,22 +4663,22 @@
       "license": "MIT"
     },
     "node_modules/@redhat-cloud-services/frontend-components": {
-      "version": "4.2.4",
-      "resolved": "https://registry.npmjs.org/@redhat-cloud-services/frontend-components/-/frontend-components-4.2.4.tgz",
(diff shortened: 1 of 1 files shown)


mkholjuraev (approved):
LGTM!

package-lock.json
```diff
@@ -4663,22 +4663,22 @@
       "license": "MIT"
     },
     "node_modules/@redhat-cloud-services/frontend-components": {
-      "version": "4.2.4",
-      "resolved": "https://registry.npmjs.org/@redhat-cloud-services/frontend-components/-/frontend-components-4.2.4.tgz",
-      "integrity": "sha512-5xne/DD1/+txtN50YOrKPeJKpSgENkGz4o4oqYMRB+ElJwnupsaOasvyfMwImKggDcAZ/uWLXt8KNoE8ZXhVvA==",
+      "version": "4.2.5",
```
bastilian, line 4666: I would suggest to save the bumped versions of direct dependencies in `package.json` as well, because that is the place where to look first, and someone might easily just drop the lock file and end up with older version.

Andrewgdewar:
Can be closed. 
Changes added [here](https://github.com/RedHatInsights/patchman-ui/pull/1206).

PR #1177, opened 2024-03-21
fix(RHINENG-8948): update request content-type for vulnerability api calls

mkholjuraev:
Updates the request content-type from application/json to application/vnd.api+json. After the vulnerability api has upgraded their connexion package, application/json type is no longer supported


AsToNlele (approved):
LGTM, saw the discussion on slack

AsToNlele:
Can't see why the PR build failed tho, I'm logged in image

PR #1178, opened 2024-04-05
chore(RHINENG-7042): adopt fec-config

mkholjuraev:
# Description

Resolves: https://issues.redhat.com/browse/RHINENG-7042

# How to test the PR

# Before the change

# After the change

# Dependent work link

# Checklist:


fec.config.js
```diff
@@ -0,0 +1,29 @@
...
+    appUrl: '/insights/patch',
+    debug: true,
+    useProxy: process.env.PROXY === 'true',
+    proxyVerbose: true,
+    interceptChromeConfig: false,
+    plugins: [],
+    ...(process.env.HOT ? { hotReload: process.env.HOT === 'true' } : { hotReload: true }),
+    ...(process.env.port ? { port: parseInt(process.env.port) } : {}),
+    moduleFederation: {
+        exclude: ['react-router-dom'],
```
bastilian, line 13: I'm very suspicious of this. Do we really need to add this to every app, even if the react router version is v6?

I think we only need to do this if the version is not v6, like policies for example.

bastilian: I mean both `expose` and `shared`.

mkholjuraev: Not all apps have upgraded to v6. Can this be a reason to have this in place still?

mkholjuraev: I understand that once all are migrated, we were expected to remove this.

bastilian: @mkholjuraev I really think we only need to do this in apps that are not migrated, and not in all till all are migrated. Chrome is sharing its version, which is v6. This would cause issues in apps with v5 and there we can add these definitions to make the app share its version instead of chrome.

bastilian: If without this the app still works proper, we probably don't need it.

mkholjuraev: I have tried it out. I have removed both exclude&shared configs related to router, and the UI does not work with this error: **useRoutes() may be used only in the context of a <Router> component.**

mkholjuraev: Call Stack:
 invariant
  vendors-node_modules_patternfly_react-component-groups_node_modules_clsx_dist_clsx_mjs-node_m-af6625.js:12071:11
 useRoutesImpl
  vendors-node_modules_patternfly_react-component-groups_node_modules_clsx_dist_clsx_mjs-node_m-af6625.js:34809:102
 useRoutes
  vendors-node_modules_patternfly_react-component-groups_node_modules_clsx_dist_clsx_mjs-node_m-af6625.js:34804:10
 Routes
  vendors-node_modules_patternfly_react-component-groups_node_modules_clsx_dist_clsx_mjs-node_m-af6625.js:35693:10
 renderWithHooks
  chrome-root.b25b7d405e94b287.js:37887:18
 mountIndeterminateComponent
  chrome-root.b25b7d405e94b287.js:41651:13
 beginWork
  chrome-root.b25b7d405e94b287.js:43164:16
 HTMLUnknownElement.callCallback
  chrome-root.b25b7d405e94b287.js:25746:14
 Object.invokeGuardedCallbackDev
  chrome-root.b25b7d405e94b287.js:25795:16
 invokeGuardedCallback
  chrome-root.b25b7d405e94b287.js:25859:31

fec.config.js
```diff
@@ -0,0 +1,28 @@
...
+    ...(process.env.HOT ? { hotReload: process.env.HOT === 'true' } : { hotReload: true }),
+    ...(process.env.port ? { port: parseInt(process.env.port) } : {}),
+    moduleFederation: {
+        shared: [
+            {
+                'react-router-dom': {
+                    singleton: true,
+                    import: false,
+                    version: '^6.8.1',
+                    requiredVersion: '^6.8.1'
```
bastilian, line 19: I am wondering what the best practice is with these version props and what would be the best to avoid running into issues. 

I'm thinking maybe they should be a bit more open. Similar to the sem ver definitions: https://github.com/RedHatInsights/frontend-components/blob/master/packages/components/package.json#L45-L46

For the `requiredVersion` we could use `>=6.0.0 <7.0.0` and I think we can drop the `version`, as webpack will default to the version in `package.json`

fec.config.js
```diff
@@ -0,0 +1,28 @@
+const { resolve } = require('path');
+
+module.exports = {
+    appUrl: '/insights/patch',
+    debug: true,
+    useProxy: process.env.PROXY === 'true',
+    proxyVerbose: true,
+    interceptChromeConfig: false,
```
bastilian, line 8: This can be omitted as `interceptChromeConfig` is checked to be explicitly set to `true`.

fec.config.js
mkholjuraev: is not `requiredVersion: '^6.8.1' `equal to `requiredVersion: ' >=6.0.0 <7.0.0'`?

bastilian: I'm not sure we need it to be exactly higher than `6.8.1`, or?

In general unless we have a very specific reason to stick to a very specific version, we should keep them "unspecific". 6.8.1 is not even the version that is in the lock file, it is `6.22.3`, or in the package.json, which is `^6.11.1`,

This leads to confusion, which is the correct version. Let's stick to `>=X.0.0` definitions. Then we can also just copy and paste these things, as they'll apply equally in each project.

fec.config.js
```diff
@@ -0,0 +1,28 @@
...
+    plugins: [],
+    ...(process.env.HOT ? { hotReload: process.env.HOT === 'true' } : { hotReload: true }),
+    ...(process.env.port ? { port: parseInt(process.env.port) } : {}),
+    moduleFederation: {
+        shared: [
+            {
+                'react-router-dom': {
+                    singleton: true,
+                    import: false,
```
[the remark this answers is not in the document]
mkholjuraev: Looking at the documentation, I believe the default value is not false for the import prop rather it is the name of the provided package. I tried removing it and I got the same error as above ^. Maybe I do not comprehend the complexity of this shared scope logic.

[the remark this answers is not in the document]
bastilian: Ah, my bad. I thought that first part below the prop title in the docs states the default.

fec.config.js
mkholjuraev: @bastilian I have made the necessary changes, PTAL

fec.config.js
```diff
@@ -0,0 +1,27 @@
...
+                    version: '^6.8.1',
```
bastilian, line 17: The `version` will default to the `package.json` version, so this can be removed.

bastilian: Oh, right, Ok. Then let's leave it for now, but I'll investigate why we have this guard in the frontend components federated modules module.

mkholjuraev: Will merge once, there is an approval

bastilian (approved):
Let's get this merged! Thank you @mkholjuraev!

PR #1181, opened 2024-04-24
refactor(TemplateWizard): Remove obsolete code

leSamo:
- removed commented in docs template link that was planned, but later scrapped
- removed dead code in `ConfigurationStepFields` which was unreachable due to `shouldShowRadioButtons = false;`, this code was only used for migrations between two development versions of Template wizard


leSamo:
@LiorKGOW I don't think we will be able to get the PR check passing since the QA tests are faulty I believe

mkholjuraev (approved):
LGTM! tested locally, no issues I encountered

PR #1183, opened 2024-05-10
fix: double icons in  display name column on package systems table

mkholjuraev:
# Description

Associated Jira ticket: No ticket needed

A small fix for the double icons on the package systems page. As the page completely uses the inventory table and also there were extra column definition, double icons appeared. Removing the extra one fixes it.

I am not sure even if this icon should be displayed in package systems table. I will consult it with the ux.

# How to test the PR

# Before the change

# After the change

# Dependent work link

# Checklist:

### DIFF
--- a/src/SmartComponents/Systems/SystemsListAssets.js
+++ b/src/SmartComponents/Systems/SystemsListAssets.js
@@ -96,14 +96,6 @@ export const advisorySystemsColumns = () => [
 
 export const packageSystemsColumns = [
     {
-        key: 'display_name',
-        title: 'Name',
-        composed: ['facts.os_release', 'display_name'],
-        props: {
-            width: 40
-        }
-    },
-    {
         key: 'tags',
         title: 'Tags',
         props: { width: 10, isStatic: true }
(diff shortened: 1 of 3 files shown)


src/SmartComponents/Systems/SystemsListAssets.js
```diff
@@ -95,14 +95,6 @@ export const advisorySystemsColumns = () => [
 ];
 
 export const packageSystemsColumns = [
-    {
-        key: 'display_name',
```
bastilian, line 99: I think we still have to have `'display_name'` as a string in the column definition. That way the Inventory knows where we want to show the display_name column, but we won't override the default column.

... I believe `{ key: 'display_name' }` might work too.

mkholjuraev: I believe we are safe to remove this column definition. Because, the inventory module [merges](https://github.com/RedHatInsights/insights-inventory-frontend/blob/97749de23e2d1851e3bb9dd6533af29fbe53e0c4/src/components/InventoryTable/hooks/useColumns.js#L41) those columns by default to those patch related columns that we pass in.

mkholjuraev: I have removed tags column definition as well as it is also not necessary

gkarat (approved):
LGTM, thanks @mkholjuraev

PR #1185, opened 2024-06-17
fix(GlobalFilters): RHINENG-6296 push correct params to url

xmicha82:
# Description

Associated Jira ticket: # [RHINENG-6296](https://issues.redhat.com/browse/RHINENG-6296)

Fixed pushing global filters to URL

# How to test the PR

1. Open Patch systems page
2. Apply Global filters
3. Global Filters should be added to the URL

# Before the change

image

# After the change

image

# Checklist:

- [x] Screenshots before and after the change are added

### DIFF
--- a/src/SmartComponents/Systems/SystemTable.test.js
+++ b/src/SmartComponents/Systems/SystemTable.test.js
@@ -27,9 +27,11 @@ const initStore = (state) => {
     return mockStore(state);
 };
 
+const apply = () => {};
+
 const renderComponent = async (mockedStore) => {
     render(<ComponentWithContext renderOptions={{ store: initStore(mockedStore) }}>
-        <Systems patchSetState={{}}/>
+        <Systems patchSetState={{}} apply={apply}/>
     </ComponentWithContext>);
 
     await waitFor(() => {
@@ -269,5 +271,9 @@ describe('SystemsTable', () => {
             }),
             {}
         );
+    });
+
+    it('should', async () => {
+
     });
 });
--- a/src/SmartComponents/Systems/SystemsTable.js
+++ b/src/SmartComponents/Systems/SystemsTable.js
@@ -14,7 +14,8 @@ import {
     arrayFromObj, persistantParams, filterSelectedActiveSystemIDs
 } from '../../Utilities/Helpers';
 import { useBulkSelectConfig, useGetEntities, useOnExport,
-    useRemoveFilter, useRemediationDataProvider, useOnSelect, ID_API_ENDPOINTS
+    useRemoveFilter, useRemediationDataProvider, useOnSelect, ID_API_ENDPOINTS,
+    usePushUrlParams
 } from '../../Utilities/hooks';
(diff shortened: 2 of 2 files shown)


src/SmartComponents/Systems/SystemTable.test.js
```diff
@@ -27,9 +27,11 @@ const initStore = (state) => {
     return mockStore(state);
 };
 
+const apply = () => {};
```
bastilian, line 30: Explain.

bastilian:
Can you elaborate a bit more on what the issue was?

src/SmartComponents/Systems/SystemTable.test.js
xmicha82: Since `apply` is now used inside the `SystemsTable` (before, it was just passed along to other functions), the tests fail if `apply` is undefined. thus i'm just passing a plain function into the component

xmicha82:
basically, `SystemsTable` did not use the hook for managing URL params which is used in other places in patch. using this hook fixed the issue

src/SmartComponents/Systems/SystemTable.test.js
bastilian: We will need to make this work even when no `apply` is provided.

bastilian:
@xmicha82 This doesn't feel like a fix, but a work around. We should also not need this `firstMount` state.

xmicha82:
@bastilian this change is consistent with other pages in patch, which also use the `historyPusher`. could you please explain why do you think it's not the right way to fix this issue?

bastilian:
@xmicha82 I took a quick look at this. I used `master` without any changes, but this in the `SystemsTable`:

```js
    const location = useLocation();
    console.log('LOCATION CHANGE', location);
```

Here is what I saw:

https://github.com/user-attachments/assets/82305bc1-6e59-402f-a000-e138fa3400a1

If you pay attention to the console you will notice that, first of all, there are probably updates to the URL multiple times per render and re renders, and if you look that location after the global filter is set you'll see that at some point the `hash` is lost.

A proper fix for this issue will require us to take a closer look and see why, how and where changes to the URL happen when the `SystemsTable` loads and ensure that all of these places do not change the `hash` and preserve the global filter. 

Ideally we would also look into reducing the amount of updates to just one update to the URL per filter change, but for now just making sure that the global filter hash is not lost should be fix enough.

bastilian:
A better place to start debugging might be the `SystemsMainContent` where we use `useSearchParams`. Adding this log output can give you an idea of how many times the URL is updated:

```js
    const setSearchParams = (...args) => {
        console.log('SetSearch', ...args);
        setSearchParamsOrig(...args);
    };
```

The `setSerachParams` function provided by react router also allows navigate options as a second parameter, this might allow us to set the `hash` of the current location, that includes the global filter.

bastilian:
@xmicha82 Have you had time to revise this issue? If you want we can debug this together and see where it starts loosing the `hash` property.

PR #1238, opened 2025-01-10
chore: Repo cleanup

leSamo:
- Remove Travis deploy step -- Remove obsolete files and constants related to Travis deploy step which is not performed since containerized builds migration
- Remove firebase, sitemap and sonarqube config files -- files inherited from repository template -- haven't been used for at least 4 years
- Remove Husky -- remove pre-commit test run, to increase consistency with other Interact repos and remove problems with performing commits through VS Code
- Remove rollup packages -- rollup is not used anymore
- Remove semantic release files -- semantic releases were removed around 2 years ago
- Remove redundant svg file inherited from template repository
- Remove old Messages file -- it hasn't been touched in the last 4 years, messages are loaded from a different file


build/messages/src/Messages.json
gkarat, line 1: This is the compiled version of intl messages (updated after running `npm run translations`, which most probably has never been done). I would suggest leaving this file or make a complete removal of translations-related scripts. Although, the app will still work and compile messages as it will default to defaultMessage's from src/Messages.js.

gkarat (approved):
Just one thing regarding intl. Overall, looks nice and can be merged.

build/messages/src/Messages.json
leSamo: Good point, I removed the translation scripts.

PR #1285, opened 2025-02-11
chore(deps): remove webpack dependancies because they are not needed

Fewwy:
# Description
According to the discussion we had on slack -> we need to remove webpack deps because we don't need it

Also, I added missing unstable spdy to fec config


bastilian (approved):
LGTM! Thank you @Fewwy!

PR #1294, opened 2025-02-24
fix Cypress coverage and collection

opacut:
Fix code coverage scripts and collection so that we have a cypress report generated for codecov.

Review steps:
- npm i
- npm run travis:verify
- check that the coverage-cypress/lcov.info is not empty
- check that the coverage-jest/lcov.info is not empty
- check that .nyc_output/out.json is not empty


LightOfHeaven1994 (approved):
LGTM! Great work!

PR #1305, opened 2025-03-07
chore: Remove old templates feature flag

leSamo:


src/Routes.js
```diff
@@ -61,21 +60,8 @@ const PackageDetail = lazy(() =>
...
-        /* webpackChunkName: "Templates" */ './SmartComponents/PatchSet/PatchSet'
-    )
-);
-
-const TemplateDetail = lazy(() =>
-    import(
-        /* webpackChunkName: "TemplateDetail" */ './SmartComponents/PatchSetDetail/PatchSetDetail'
-    )
-);
-
```
gkarat, line 75: Can we remove these components since they are no longer used in patch?

src/SmartComponents/Systems/SystemsTable.js
```diff
@@ -188,7 +181,7 @@ const SystemsTable = ({
                 actionResolver: (row) =>
                     systemsRowActions(
                         activateRemediationModal,
-                        !templateUpdateEnabled  && openAssignSystemsModal,
+                        false,
```
gkarat, line 184: nit-pick: we could do even more clean-up here by removing the second argument completely from the function and always return only one action "Apply all applicable advisories" from systemsRowActions

src/SmartComponents/SystemDetail/InventoryDetail.js
```diff
@@ -94,34 +83,20 @@ const InventoryDetail = () => {
...
-                            isDisabled: !hasTemplateAccess || satelliteManaged,
-                            onClick: () => openAssignSystemsModal({ [inventoryId]: true })
-                        },
-                        {
-                            title: intl.formatMessage(messages.titlesTemplateRemoveMultipleButton),
-                            key: 'remove-from-template',
-                            isDisabled: !hasTemplateAccess || !patchSetName || satelliteManaged,
-                            onClick: () => openUnassignSystemsModal([inventoryId])
-                        }]}
+                    actions={[]}
```
gkarat, line 86:
```suggestion
```

src/Routes.js
leSamo: Yes, this will come in a subsequent PR, I didn't want this one to be so bloated.

src/SmartComponents/Systems/SystemsTable.js
leSamo: Yes, this will come in a subsequent PR, I didn't want this one to be so bloated.

gkarat (approved):
LGTM, thanks @leSamo

PR #1310, opened 2025-03-24
feat(RHINENG-15598): Add column management to Patch InventoryTable tables

leSamo:
https://issues.redhat.com/browse/RHINENG-15598

### DIFF
--- a/src/SmartComponents/AdvisorySystems/AdvisorySystemsTable.js
+++ b/src/SmartComponents/AdvisorySystems/AdvisorySystemsTable.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Fragment } from 'react';
 import { TableVariant } from '@patternfly/react-table';
 import { InventoryTable } from '@redhat-cloud-services/frontend-components/Inventory';
 import propTypes from 'prop-types';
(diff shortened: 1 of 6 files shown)


bastilian:
@leSamo Would it not make more sense to implement the column management directly in the `InventoryTable` component, instead of implementing it individually in each application?

leSamo:
@bastilian yes, that would be an option, but honestly that would only save a few lines of code here. I think that Inventory in its current state is crumbling under its own weight and we should be careful with adding new features. Future of Inventory is also quite uncertain (Inventory-centric views?).

leSamo:
@bastilian Created a ticket for it: https://issues.redhat.com/browse/RHINENG-16902

johnsonm325:
Code looks good. If you remove all columns but one in the Advisories table, the expandable and selection columns get very wide. It does not do it on the Systems table. Maybe this is easy to fix?

image

bastilian:
> @bastilian Created a ticket for it: https://issues.redhat.com/browse/RHINENG-16902

That's great! We should now implement the column management in InventoryTable first and then  enable it for the apps using the InventoryTable, without having to implement the management in those apps themselves for their `SystemsTable`s

PR #1322, opened 2025-04-25
feat(RHINENG-15545): Consume last seen column from Inventory

leSamo:
https://issues.redhat.com/browse/RHINENG-15545

The tooltip was added to the Inventory last seen column, but we were overwriting the column with our title due to intl, but it's not needed anymore. The column header will have tooltip only after RedHatInsights/insights-inventory-frontend#2429 has been merged.


bastilian (approved):
LGTM! Thank you @leSamo!

PR #1424, opened 2025-10-29
chore: Update dependabot to ignore jest, pf and rh cloud services

avitova:
# Description

I copied the settings of [Dependabot](https://github.com/content-services/content-sources-frontend/blob/main/.github/dependabot.yml) for Content. The patternfly, react-router dom and rh cloud services use higher versions. I think this should also be updated in content, as you already migrated pf, react-router-dom and rh services to higher versions than you ignore.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,9 +6,31 @@ updates:
       interval: "weekly"
     labels:
       - "dependencies"
-    open-pull-requests-limit: 5
     commit-message:
       prefix: "chore(deps):"
+    ignore:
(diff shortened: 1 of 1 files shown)


marusak:
In Content we also have https://github.com/content-services/content-sources-frontend/blob/main/.github/dependabot.yml#L35 which makes sure, we get one PR for everything and not bunch of small ones. Should we also add something like that?

LightOfHeaven1994 (approved):
LGTM, there is an issue with commit lint so `test` pipeline fails. Can you fix that? 🙏

TenSt (approved):

PR #1457, opened 2025-12-19
HMS-9725: Add test to verify exporting feature on patch

mayurilahane:
# Description

- Export data into CSV and JSON.
- Test exporting ability on the "Advisories", "Packages" 
and "Systems" pages.
- Validate the data structure and some of the contents.
- Used "papaparse" to parse the CSV files and 
added it to the dependencies.
- Fixed advisories navigation to use exact matching. -

https://issues.redhat.com/browse/HMS-9725

### DIFF
--- a/playwright/UI/VerifyExportingFeature.spec.ts
+++ b/playwright/UI/VerifyExportingFeature.spec.ts
@@ -1,0 +1,203 @@
+import * as fs from 'fs';
+import { Page } from '@playwright/test';
+import Papa from 'papaparse';
+import {
+  test,
+  expect,
+  closePopupsIfExist,
+  waitForTableLoad,
+  navigateToAdvisories,
+  getRowCellByHeader,
+  navigateToSystems,
+  navigateToPackages,
+} from 'test-utils';
+import { getTotalCountFromPagination } from 'test-utils/helpers/pagination';
+
+async function exportCsvJson(
+  page: Page,
+): Promise<{ csvData: Record<string, string>[]; jsonData: Record<string, string>[] }> {
+  await page.getByRole('button', { name: 'Export', exact: true }).click();
+
+  // Export to CSV
+  const [downloadCsv] = await Promise.all([
+    page.waitForEvent('download'),
+    page.getByRole('menuitem', { name: 'Export to CSV' }).click(),
+  ]);
+  const pathCsv = await downloadCsv.path();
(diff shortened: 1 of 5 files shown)


playwright/UI/VerifyExportingFeature.spec.ts
```diff
@@ -0,0 +1,214 @@
...
+  test('Verify exporting via CSV and JSON on patch', async ({ page }) => {
+    await closePopupsIfExist(page);
+
+    await test.step('Verify exporting on Advisory page', async () => {
+      await navigateToAdvisories(page);
+      await waitForTableLoad(page);
+      const page_count = await getTotalCountFromPagination(page);
+      // target all repos and then select the first one
+      const repos = page.locator('tbody[role="rowgroup"]');
+      const firstRepo = repos.first();
```
dominikvagner: the test is trying to export advisories, not repos, so the naming should reflect that 😅

playwright/UI/VerifyExportingFeature.spec.ts
```diff
@@ -0,0 +1,214 @@
...
+      const [nameText, synopsisText, typeText] = await Promise.all([
+        nameCell.textContent(),
+        synopsisCell.textContent(),
+        typeCell.textContent(),
+      ]);
+
+      const { csv, json } = await exportCsvJson(page);
+
+      // Parse CSV and find matching row by id (which matches UI name)
+      const csvResult = Papa.parse<Record<string, string>>(csv, {
```
dominikvagner: I think the parsing could be done in the export helper since it's repeated in every time after the export helper is called 💭

playwright/UI/VerifyExportingFeature.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+
+  return { csvData: csvResult.data, jsonData: JSON.parse(json) };
+}
+
+test.describe('Verify exporting feature', () => {
+  test('Verify exporting on Advisory page', async ({ page, systems }) => {
+    await systems.add('export-advisory-test', 'base');
+    await closePopupsIfExist(page);
+    await navigateToAdvisories(page);
+    await waitForTableLoad(page);
```
dominikvagner: this `waitForTableLoad` could be probably just added to all the `navigateTo...` functions so we don't forget to use it down the line 😅

PR #1470, opened 2026-01-15
HMS-9728: add integration test for template assignment

katarinazaprazna:
# Description

I’ve added a new integration test, `CanAssignTemplateToSystem.spec.ts`, to cover the end-to-end template assignment flow. This walks through registering a system with a unique identity, creating a patch template, and assigning it, finally verifying that the Systems UI updates the "Template" column correctly.

I introduced a `cleanupTemplates` helper to ensure test data is properly removed and updated the cleanup fixture to prevent premature teardown. I also added a `disableTrackingAndConsent` utility to block external scripts like TrustArc and Amplitude, which helps reduce console noise.

Associated Jira ticket: https://issues.redhat.com/browse/HMS-9728

## Other

Added a condition to track failed task statuses. This prevents infinite loops and test hangs by ensuring the polling mechanism recognizes both success and failure as terminal states. This addresses observed issues where failing tasks caused the test suite to hang indefinitely.

I ended up using the Patch API for template assignment rather than `rhc`. I found that `rhc` wouldn't apply the template if it detected the system was already registered, which forced me to disconnect and reconnect the system just to make it work. Switching to the API for assignment (while keeping `rhc` for registration) was much cleaner, though it took a bit of refactoring to get right.

# How to test the PR

1. Run the specific test and ensure it passes successfully
2. Verify that no stale templates with the prefix `TemplateAssignmentTest-` remain in the environment after the test completes

# Checklist:

- [x] Tests for the changes have been added


playwright/Integration/CanAssignTemplateToSystem.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+
+    await test.step('Boot and register system', async () => {
+      await client.Boot('rhel9');
+
+      const registration = await client.RegisterRHC();
+      if (registration?.exitCode !== 0) {
+        throw new Error(`System registration failed with exit code: ${registration?.exitCode}`);
+      }
+
+      await waitForRhcdActive(client);
```
marusak: FWIW, these two should not be needed.

katarinazaprazna: Thanks for catching this. It's an artifact from the `rhc` approach to template assignment. I'll clean it up!

playwright/Integration/CanAssignTemplateToSystem.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+    await test.step('Verify template assignment in API', async () => {
+      const isAttached = await pollForSystemTemplateAttachment(page, hostname, true, 5_000, 6);
+      expect(
+        isAttached,
+        `Template '${templateName}' should be attached to system '${hostname}'`,
+      ).toBe(true);
+    });
+
+    await test.step('Verify template is displayed in Systems table', async () => {
+      await navigateToSystems(page);
```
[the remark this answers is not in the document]
katarinazaprazna: I was operating under the assumption that the banner(s) would reappear on every navigation or page refresh. I will verify the expected behaviour

PR #1489, opened 2026-02-05
test: stabilize advisory remediation plan test

xbhouse:
# Description

- Updates the advisory remediation plan test to 
  - only check for the relevant system in the remediation plan 
  - use a separate user to avoid conflicts with other systems
- Refactors the system fixture to allow specifying which user/token to create a system for
- Refactors the test setup/authentication to include [these enhancements](https://github.com/content-services/content-sources-frontend/pull/829)

# How to test the PR

All tests pass

# Before the change

# After the change

# Dependent work link

# Checklist:

### DIFF
--- a/.github/workflows/ui-tests.yml
+++ b/.github/workflows/ui-tests.yml
@@ -75,7 +75,7 @@ jobs:
       - name: Increase file watchers limit
         run: echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
 
-      - name: Install dependecies
+      - name: Install dependencies
         run: 'npm ci && npm run build && npx playwright install chromium --only-shell'
 
       - name: Update /etc/hosts
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -56,7 +56,7 @@ export default defineConfig({
           name: 'Integration',
           use: {
             ...devices['Desktop Chrome'],
-            storageState: '.auth/admin_user.json',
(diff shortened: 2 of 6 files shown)


xbhouse:
i **think** the advisory remediation plan test is failing because the set of systems the plan applies to isn't always stable :thinking: thinking through this a bit more

dominikvagner:
> yep that makes sense, i'll fix that. there's another issue too where the plan creation fails completely if systems associated with that advisory aren't available in inventory (i guess because they're removed after another test finishes):
> 
> ```
> {
>   "errors": [
>     {
>       "id": "4aeda25b790e4a5d8bca4ad7f7dbef1a",
>       "status": 400,
>       "code": "UNKNOWN_SYSTEM",
>       "title": "One or more requested systems do not exist in Inventory"
>     }
>   ]
> }
> ```
> 
> i'm thinking we could solve this by either removing all systems only at the end of the test suite, or by using a separate user for the advisories tests. i'm leaning towards using a separate user, a global teardown might be a bit more complicated. what do you think @dominikvagner ?

oh, didn't see that in the logs at first 🔍
 
yeah, global teardown would mean we would have to change how that entire systems fixture works now so I wouldn't go with that :/
I would either use a separate user, or add a tag for tests we can't/don't want to run in parallel and then run them in a separate project right after all the other UI tests (tags and grep), that might be the simplest solution 💭😅 

_(before we start adding more users we would also want to tweak how we do auth here and use the improved way how we do it in content sources, I could help with that if wanted 😄)_

PR #1582, opened 2026-04-16
Test: add prefix to test names

swadeley:
# Description

Adding a prefix and using the name returned by the system fixture for test names.
To ensure a system found is the one created in that specific test run.

### DIFF
--- a/playwright/UI/FilterTypePatch.spec.ts
+++ b/playwright/UI/FilterTypePatch.spec.ts
@@ -16,6 +16,8 @@ import {
   getWorkspaceGroup,
 } from 'test-utils';
 
+const FILTER_TEST_PREFIX = 'filter-test';
+
 /** One day in ms; used for date-only display tolerance. */
 const DAY_MS = 24 * 60 * 60 * 1000;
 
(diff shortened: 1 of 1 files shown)


playwright/UI/FilterTypePatch.spec.ts
```diff
@@ -323,11 +331,11 @@ test('Verify filter contents', async ({ page, systems }) => {
     await expect(
       page
         .locator('td[data-label="Name"]')
-        .filter({ hasText: 'filter-contents-test-base' })
+        .filter({ hasText: filterContentsBaseName })
         .first(),
     ).toBeVisible();
     await expect(page.locator('td[data-label="Name"]').first()).toContainText(
-      'filter-contents-test-base',
+      filterContentsBaseName,
```
xbhouse: i was thinking we could just check for `system.name` in the table, which would be the name of the system we get from the response after creating a system. 

you could also set that to a clearer variable name so it's easier to see which system we're expecting?

swadeley: e.g. baseSystem.name

xbhouse (approved):
thank you!! tests pass locally (with the workspace test case commented out). ack 🚀

PR #1587, opened 2026-04-21
chore: update readme and rules for agents

xbhouse:
# Description

- Adds initial setup for agents with the same config from [content-sources-frontend](https://github.com/content-services/content-sources-frontend/pull/963). Moves the existing Playwright rule so it's auto-loaded by  both Claude and Cursor
- Adds a rule and some initial guidelines to help agents help us with migrating to TypeScript
- Updates README

# Checklist:

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,17 @@
+# AGENTS.md
+
+Project: Frontend for Red Hat's Patch service (React + JavaScript + TypeScript + PatternFly React)
+
+Backend: https://github.com/RedHatInsights/patchman-engine
+
+## Dev environment setup
+- Use `nvm use` to match node version in `.nvmrc`
+- Run [script to patch your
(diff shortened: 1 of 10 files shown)


AGENTS.md
```diff
@@ -0,0 +1,15 @@
+# AGENTS.md
+
+Project: Frontend for Red Hat's Patch service (React + JavaScript + TypeScript + PatternFly React)
```
katarinazaprazna, line 3: I'm wondering if we should also let the agents know where [the backend](https://github.com/katarinazaprazna/patchman-engine) lives. It may be helpful for the TS migration as well

katarinazaprazna (dismissed):
Thanks for the changes!

I left one small suggestion. Feel free to consider it if you think it adds value, but otherwise, awesome work on this :)

AGENTS.md
xbhouse: oh yes :) thanks! just added that

katarinazaprazna (approved):
Thank you! ❤️

PR #1588, opened 2026-04-22
chore: add analytics API keys to frontend deploy config

dajohnso:
## Summary
- Adds the `analytics` block with `APIKey` and `APIKeyDev` to the `module` section in `deploy/frontend.yml`
- Aligns with the pattern used by other frontends (inventory, advisor, tasks, ocp-advisor)
- Fixes RHINENG-25987
## Test plan
- [ ] Verify the YAML is valid and the deploy CRD parses correctly
- [ ] Confirm analytics events are sent in stage with the correct API key

### DIFF
--- a/deploy/frontend.yml
+++ b/deploy/frontend.yml
@@ -22,7 +22,14 @@ objects:
           - /apps/patch
       image: ${IMAGE}:${IMAGE_TAG}
       module:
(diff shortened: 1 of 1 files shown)


swadeley (approved):
ACK

PR #1592, opened 2026-04-29
fix(RHINENG-26137): deal with permissions correctly based on FF

LightOfHeaven1994:
This PR deals with fetching permissions correctly. It will avoid calling both rbac & kessel hooks at the same time and it will be done based on feature flag value.

Also we are properly waiting until feature flag is fetched, for that we use `flagsReady`.

### How to test:

1. Navigate to `Advisories` page for example 
2. Open devtools network tab
3. Refresh the page and check if there are no RBACv1 requests made for patch `access/?application=patch`
4. Check that app loads as usual


xbhouse (approved):
lgtm! thank you ❤️

dominikvagner (approved):
ack! ✅ thx 👏🏼 

_ahh, Bryttanie beat me to it 🐌😄_

PR #1596, opened 2026-04-30
chore: increase title length to 70

rverdile:
I noticed most of the commits here fail the test for title length, and it gets ignored and merged anyway. Let's increase it to a number that might be more convenient to enforce.


xbhouse (approved):
thank you!! 🚀

PR #1598, opened 2026-05-01
fix: handle empty relations and reduce test flakiness

xbhouse:
# Description

- Adjusts permission check to wait for relations to load fully before granting/denying access
- Updates playwright UI test config and AdvisoriesTests to reduce flakiness 
  - First issue: When multiple users fetch workspaces at the same time, this seems to cause auth leakage across sessions so workspaces for the wrong user might be fetched to determine access. To prevent this, this PR isolates UI test projects by user and runs each project serially
  - Second issue: Since AdvisoriesTests creates a remediation plan for an advisory, any systems the advisory is associated with are included in the request to remediations. If a system in the request has already been deleted in inventory, it will still be in patch for around 10 minutes. Retrying will almost always result in this kind of failure, so this PR disables retries for this test
 
# How to test the PR

- The "not authorized" page shouldn't flash briefly on initial app load or refresh
- There shouldn't be any test failures related to rejected access

# Checklist:

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -29,10 +29,10 @@ export default defineConfig({
     launchOptions: {
       args: ['--use-fake-device-for-media-stream'],
     },
-    ...(process.env.TOKEN
+    ...(process.env.ADMIN_TOKEN
       ? {
           extraHTTPHeaders: {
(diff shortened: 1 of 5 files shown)


swadeley (approved):
ACK

PR #1641, opened 2026-05-27
RHINENG-25944: ensure correct encoding of tag filters

dominikvagner:
## Summary
This makes sure the global tag filters are correctly URI encoded, only a
single encode pass. Previously they could be double encoded which made
the API always return no systems.

Associated Jira ticket: [#RHINENG-25944](https://redhat.atlassian.net/browse/RHINENG-25944)

> Also includes a QOL fix for local development, _that sets the maxWorkers 
> in the Jest config to prevent accidental computer lock-ups/freezes when running
> tests, because the default is using the maximum of available resources._

## Testing steps
Test out filtering with both the "local" tag filters (in the table toolbar) and with the global filters at the top.
Both of them should now work.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -16,4 +16,5 @@ module.exports = {
   transformIgnorePatterns: [
     '<rootDir>/node_modules/(?!(@patternfly/react-core/|@patternfly/react-icons/|@redhat-cloud-services|@openshift|lodash-es|@patternfly/react-table|@patternfly/react-tokens|p-all)).*$',
   ],
+  maxWorkers: 4,
 };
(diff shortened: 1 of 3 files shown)


TenSt (approved):
lgtm

xbhouse:
FYI the "Advisories Tests › Create a new remediation plan" UI test will likely fail until culled systems can be deleted in patch again. this [PR](https://github.com/RedHatInsights/patchman-engine/pull/2229) should fix that :crossed_fingers:

jest.config.js
```diff
@@ -16,4 +16,5 @@ module.exports = {
   transformIgnorePatterns: [
     '<rootDir>/node_modules/(?!(@patternfly/react-core/|@patternfly/react-icons/|@redhat-cloud-services|@openshift|lodash-es|@patternfly/react-table|@patternfly/react-tokens|p-all)).*$',
   ],
+  maxWorkers: 4,
```
xbhouse, line 19: yes!! thank you 🙌

PR #1645, opened 2026-05-27
test: reduce filter test runtime

xbhouse:
# Description

Removes unnecessary waits from the filter test helpers. This should reduce the runtime of the  "Patch Filters" > "Filter types on Advisories page" UI test.  

# How to test the PR

UI filter tests shouldn't time out

### DIFF
--- a/playwright/UI/FilterTypePatch.spec.ts
+++ b/playwright/UI/FilterTypePatch.spec.ts
@@ -297,8 +297,6 @@ test('Verify filter contents', async ({ page, systems }) => {
 
   await navigateToSystems(page);
   await closePopupsIfExist(page);
-  await openConditionalFilter(page);
-  await resetFilters(page);
 
   await test.step('Verify "Operating system" filter contents', async () => {
     await openConditionalFilter(page);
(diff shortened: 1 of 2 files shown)


swadeley (approved):
ACK

swadeley:
Hi @xbhouse , " UI/FilterTypePatch" still fails but that is because the base system which is in the Workspace called "TestSpace" is not in the results, in the table. But it is there when the test is still at the filter by operating stage of the test.

xbhouse:
> Hi @xbhouse , " UI/FilterTypePatch" still fails but that is because the base system which is in the Workspace called "TestSpace" is not in the results, in the table. But it is there when the test is still at the filter by operating stage of the test.

that's expected @swadeley, the workspace filtering should be fixed once Inventory reverts the breaking changes. this PR only affects the "Filter types on Advisories page" test (reduces the runtime of that test from about 4 minutes to 1.5 minutes)  and we shouldn't see timeouts like [this](https://github.com/RedHatInsights/patchman-ui/actions/runs/26517766967/job/78130640145) anymore

